PR #270, opened 2016-07-03
Fix search index preparation for departments

wetneb:
@Lysxia : I had a problem with search index population for departments: not all researchers have departments, so the original code failed when `r.department == None`. But anyway why don't we simply index all departments?


wetneb:
@Lysxia has a nicer solution :-)

PR #286, opened 2016-07-07
Separate download button from publisher

### LINKED ISSUE #171: Remove the editor in the publications lists
"surtout que là il est à côté du bouton "download" et ça sous-entend pour les gens qu'on le télécharge depuis ou grace à l'éditeur du coup"

Lysxia:
Resolves #171

Now looks like this: 
sc

### DIFF
--- a/papers/static/style/style.css
+++ b/papers/static/style/style.css
@@ -371,6 +371,13 @@ li
     color: #272743;
 }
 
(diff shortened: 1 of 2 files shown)


wetneb:
I think it's perfect. Let's leave it open for a bit so that others can have their say on it.

PR #311, opened 2016-08-11
Move to Django 1.10

wetneb:
Moving to Django 1.10 will enable us to use non-atomic migrations, which is useful for data migrations running on millions of records. This PR fixes the various bugs introduced by depreciations.


wetneb:
We should probably wait for https://github.com/jmcclell/django-bootstrap-pagination/pull/53 to be merged and released before merging this.

PR #314, opened 2016-08-13
remove hidden dependency to redis in HalTest

### LINKED ISSUE #313: make PrefilledTest independent of haystack
I'm working on the HAL binding (see #312).
when running test `deposit.hal` (`python manage.py test deposit.hal`), it fails because PrefilledTest depends on haystack and a server at `http://localhost:9200/`.

In order to simplify installation and testing, is it possible to replace the dependency to a haystack server in PrefilledTest, and to replace it for instance by loading the data from disk directly?

monperrus:
closes #313.

now one can simply run `python manage.py test deposit.hal` after a simple virtualenv/pip setup.

### DIFF
--- a/deposit/hal/protocol.py
+++ b/deposit/hal/protocol.py
@@ -77,8 +77,12 @@ class HALProtocol(RepositoryProtocol):
         # Abstract
         if self.paper.abstract:
             data['abstract'] = kill_html(self.paper.abstract)
-        else:
-            self.paper.consolidate_metadata(wait=False)
+        # commenting this else branch because
+        # it introduces a runtime dependency to the redis database
+        # however, it is good to be able to run the HAL tests
+        # with a simple django config
+        #else:
+            # self.paper.consolidate_metadata(wait=False)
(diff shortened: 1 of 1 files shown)


wetneb:
Thanks for the contribution! But… this actually disables our code, and not just for the tests! This is a bit radical, don't you think? If you want to get rid of redis and celery, the clean way to do that is to put `CELERY_ALWAYS_EAGER = True` in your settings file (for instance `dissemin/settings/__init__.py`) so that all asynchronous tasks are run synchronously, in the same thread.

wetneb:
See commit 0fd3c067c500b9ddadc4b89c11baa0ff873ac333 for a description of this setting.

monperrus:
> the clean way to do that is to put |CELERY_ALWAYS_EAGER = True| in
> your settings file (for instance |dissemin/settings/__init__.py|)
this solution works, thanks!

monperrus:
Thanks, I'm run to run

PR #365, opened 2017-03-15
Fix a few easy issues

Phyks:


wetneb:
Merging because it's obviously fine, but we need to solve this problem about foreign PRs and travis secure environment variables.

PR #368, opened 2017-03-20
Add a more explicit error message when paper is already in HAL

### LINKED ISSUE #356: More explicit error message when paper is already in HAL [$10]
Deposit record: 515 (in prod)
We shouldn't just say "HAL refused the deposit (HTTP error 400)" but explain that the paper is already in HAL.

This should be solved by modifying the code that submits the deposit to HAL:
https://github.com/dissemin/dissemin/blob/master/deposit/hal/protocol.py

<bountysource-plugin>

---
There is a **[$10 open bounty](https://www.bountysource.com/issues/42647393-more-explicit-error-message-when-paper-is-already-in-hal?utm_campaign=plugin&utm_content=tracker%2F6475570&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F6475570&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Phyks:
And more generally, always try to return a verbose error message when HAL returns a 400 error code.

Closes #356.

P.S.: I am not really sure how I could test it :/ So this is untested for the moment.

### DIFF
--- a/deposit/hal/protocol.py
+++ b/deposit/hal/protocol.py
@@ -169,15 +169,37 @@ class HALProtocol(RepositoryProtocol):
 
             xml_response = resp.read()
             conn.close()
-            if resp.status != 201:
-                self.log('Deposit response status: HTTP %d' % resp.status)
-                self.log(xml_response.decode('utf-8'))
-                raise DepositError(
-                    __('HAL refused the deposit (HTTP error %d)') % resp.status)
-
             try:
                 parser = etree.XMLParser(encoding='utf-8')
                 receipt = etree.parse(BytesIO(xml_response), parser)
+                if resp.status != 201:
+                    self.log('Deposit response status: HTTP %d' % resp.status)
+                    self.log(xml_response.decode('utf-8'))
+                    # Get the verbose description of the error to output it as well
+                    root = receipt.getroot()
+                    verboseDescription = (
+                        next(
+                            root.iter(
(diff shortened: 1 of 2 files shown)


wetneb:
@Phyks: Travis says: `./deposit/hal/protocol.py:190: undefined name 'json'` . Can you `import json` somewhere?

wetneb:
You can use the following unit test:
```python
    def test_paper_already_in_hal_but_not_in_dissemin(self):
        """
        In this case, Dissemin missed the paper on HAL
        (for some reason) and so the deposit interface was
        enabled. But HAL refuses the deposit! We have to
        give a good error message to the user.
        """
        # this paper is currently in HAL-preprod
        p = Paper.create_by_doi('10.1051/jphys:01975003607-8060700')

        # this is just to make sure that we are depositing with
        # a single author (otherwise, the deposit would fail because
        # we are not providing enough affiliations).
        p.authors_list = [p.authors_list[0]]

        r = self.dry_deposit(p,
            abstract='this is an abstract',
            topic='INFO',
            depositing_author=0,
            affiliation=59704) # ENS

        # Deposit fails: a duplicate is found
        self.assertEqualOrLog(r.status, 'failed')

        # The error message should be specific
        self.assertTrue('already in HAL' in r.message)
```

PR #420, opened 2017-07-14
[enh] replace dx.doi.org by doi.org

jibe-b:
Doi.org asks to systematically substitute doi.org to dx.doi.org

Source: https://hyp.is/M9Lj0GjaEeeB4idO26hqow/www.doi.org/factsheets/DOIProxy.html


wetneb:
Thanks a lot! That was needed indeed. Do you think we could also switch to HTTPS in the same go?

jibe-b:
So, that makes two small changes for today :) I love your reactivity!

PR #421, opened 2017-07-14
[fix] change OAB URL to stable and https

jibe-b:
See https://github.com/OAButton/discussion/issues/790


jibe-b:
@wetneb Are the papers requested through the OpenAccessButton recorded?

wetneb:
@jibe-b No we do not have any tracking at all (no analytics either btw).

jibe-b:
Great thing :)

jibe-b:
I asked because until now the request couldn't be succesful as the redirection was to the dev instance of the OAB. So there might be few one that have been started but failed.

Too late ;)

wetneb:
yeah it would be a good reason to actually have some sort of tracking, I suppose…

PR #426, opened 2017-08-02
Add OSF interface

stephno:
It is now possible to deposit papers on OSF Preprints _via_ https://api.osf.io/v2/


wetneb:
Even the test coverage increased! O_O

PLMir:
🙌🙌🙌 Formidable !
Excellent travail 👌
👍👍 
👏👏
🐤 Félicitations !

stephno:
@wetneb Yes! But I think there’s still something to do with it. It is possible that it doesn’t handle properly `test_form_with_no_subject`, regarding the problem we’ve got lately. Nevertheless, it doesn’t prevent anybody from depositing a paper, so that’s cool. :)

wetneb:
Yes that is something I'll investigate before we push it on prod - but that should not be a problem for testing.

wetneb:
Let's merge this!!

Issue #460, opened 2017-10-06
Invalid HTML on https://dissem.in/faq


a3nm:
https://validator.w3.org/nu/?doc=https%3A%2F%2Fdissem.in%2Ffaq

Some of these may be genuine glitches.

a3nm:
OK now it's really fixed...

PR #462, opened 2017-10-08
skeleton.html: Added closing div on wrapper element

NikolaJankovic:
Issue #460 - the paragraph tags are valid but the closing div was missing.

### DIFF
--- a/templates/dissemin/skeleton.html
+++ b/templates/dissemin/skeleton.html
@@ -99,6 +99,7 @@ $.ajax({ url: 'https://{{ ORCID_BASE_DOMAIN }}/userStatus.json?logUserOut=true',
                 </div>
             </nav>
         </div>
+        </div>
         <div id="content">
             {% if breadcrumbs %}
                 {% include "dissemin/breadcrumbs.html" %}


a3nm:
Hi there! Thanks a lot for trying to fix this bug. :) I had a look and I don't think this closing tag should be added to skeleton.html. I think it should be in faq.html and others. I have added it, but now other issues are popping up...

NikolaJankovic:
Hey, the issue is popping up in faq.html but only because it inherits from skeleton.html where the element with the wrapper id is located. 

Can you let me know which new issues are arising?

a3nm:
I think the div with id "wrapper" is correctly closed in the skeleton, despite the misguiding message from the W3C validator. The div that is not closed is in faq.html (and other files), with id "actualContent".

The new issues are caused by lists in faq.html that are in "p" tags and should be outside of them. I'm working on it now.

a3nm:
OK, I was able to fix these problems, and now the page [is valid](https://validator.w3.org/nu/?doc=https%3A%2F%2Fdissem.in%2Ffaq).

Thanks again for pushing us to fix this bug! I'm very sorry that we couldn't merge your pull request because the closing tag had to be added at another place. Please feel free to fix any other markup glitches if you find some.

wetneb:
@NikolaJankovic Thank you so much for your PR!

PR #482, opened 2017-10-27
always use '_' instead of '__' for localization

a3nm:
the strings with '__' are not correctly picked up

(Thanks to Chinese, it's now very easy to notice these strings!)


a3nm:
Thanks!

-- 
Antoine Amarilli

PR #490, opened 2017-11-09
Backend cleanup for oadoi

wetneb:
First part of the work to kill proaixy and switch to Crossref + OAdoi only.


eocarragain:
+1 As a downstream consumer of dissemin, great to have clarity on the underlying sources in use, so consolidating on Crossref + OADOI makes sense

wetneb:
@eocarragain it's going to take a while before dissemin is fully in sync with both sources, but it's in progress!

eocarragain:
@wetneb sure. Does this mean that awareness of articles available through researchgate will be going away from dissemin?

wetneb:
No, their records will stay, but the preferred pdf_url of papers will likely change to urls obtained from OAdoi.

eocarragain:
Thanks!

PR #491, opened 2017-11-10
Move to Django 1.11

wetneb:
This does not break compatibility with Django 1.10, it's just nice to have support for server-side cursors in our postgres connection.


wetneb:
Merging even if Travis is red (it's only a Zenodo temporary issue).

PR #497, opened 2017-11-13
Fix language on root HTML element. Closes #464

### LINKED ISSUE #464: Language on root html element is incorrect on localized versions
On https://dissem.in/faq with the French locale, for instance, the root element says `<html lang="en">`. This should be `<html lang="fr">`.

This comes from `templates/dissemin/skeleton.html`, and should be modified to reflect the actual language of the localization.

wetneb:
Adapts the language on the root element.


wetneb:
Tests are only failing because of 500 errors at OSF.

PR #513, opened 2018-01-09
Add new language

Golls:
Add new language in main page (Change language list).


wetneb:
@Golls thank you very much!

Issue #530, opened 2018-06-07
Footer on small screens


Phyks:
Hi,

I feel like there should be the following rules on small screens:

* `div#footer > div.col-xs-4 > ul` should not have any left padding.
* `div#change-lang-form-div` should be a `col-xs-6 col-xs-offset-3` (rather than a 4 and 2).

Additionnally, the `<h5>` about changing the language could be made a `<label>` of the `<select>` I think.

Before was like:

![capture d ecran 2018-06-07 a 14 43 50](https://user-images.githubusercontent.com/3856586/41100275-36b43afc-6a61-11e8-84da-d494758511b1.png)

After these changes, it would look like

![capture d ecran 2018-06-07 a 14 43 39](https://user-images.githubusercontent.com/3856586/41100291-3e281ba0-6a61-11e8-8d3d-c2f51492aded.png)

I can make a PR if you agree with these :)

PR #533, opened 2018-07-30
Footer on small screens

Phyks:
Hi,

Here is a PR matching https://github.com/dissemin/dissemin/issues/530. I'm not super used to editing Dissemin files, I hope I did it right.

First commit is the deletion of left padding in footer and rework of language switch form, as written in https://github.com/dissemin/dissemin/issues/530#issue-330255781.

Second commit is optional and expands the list in the footer across the full width on `xs` screens. Actually, even with the first commit, there were some issues left on small screens (mobile phones for instance). This is a way to fix them.

### DIFF
--- a/front/css/style.css
+++ b/front/css/style.css
@@ -244,3 +244,7 @@ h1
 {
     color: black;
 }
+
+#footer ul {
+    padding-left: 0;
+}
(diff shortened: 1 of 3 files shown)


wetneb:
Deployed!

Issue #537, opened 2018-09-23
Update OSF connector to latest API


wetneb:
OSF deposits fail, probably because OSF made changes to their API: a new field seems mandatory.

```
### What are the subjects?
[[u'584240da54be81056cecaab0']]
### Creating the metadata
{
    "data": {
        "attributes": {
            "category": "project", 
            "public": "false", 
            "tags": [
                "test"
            ], 
            "description": "this is a test abstract", 
            "title": "XVI. On two new Genera of Land Tortoises."
        }, 
        "type": "nodes"
    }
}
[
    {
        "name": {
            "last": "Bell", 
            "first": "Thomas"
        }
    }
]
--- Request to https://api.osf.io/v2/nodes/

Status code: 201 (expected 201)

### Creating a new deposition
--- Request to https://api.osf.io/v2/nodes/x7f4t/files/

Status code: 200 (expected 200)

### Uploading the PDF
--- Request to https://files.osf.io/v1/resources/x7f4t/providers/osfstorage/?kind=file&name=article.pdf

Status code: 201 (expected 201)

--- Request to https://api.osf.io/v2/nodes/x7f4t/contributors/

Status code: 201 (expected 201)

### License Debug
Requested License ID: 563c1cf88c5e4a3877f9e965
--- Request to https://api.osf.io/v2/nodes/x7f4t/

Status code: 200 (expected 200)

### Updating License
200
[...]

PR #538, opened 2018-09-23
Specify required type field in "preprints" OSF API

nemobis:
Per https://developer.osf.io/#operation/preprints_create

https://github.com/dissemin/dissemin/issues/537


wetneb:
Thanks a lot!

PR #546, opened 2019-01-07
Fix Vagrant provisioning script

Phyks:
Hi,

Here is a quick and dirty fix for the failures of Vagrant provisioning.

* I'm using `set -e` to stop the script at the first error and `set -x` to have a verbose output and ease debugging (prints the command ran before running it).
* Some `gdal` libraries were missing, these are now installed through `apt`.
* There was a weird bug with the `setuptools` install which I solved by first uninstalling `setuptools` and reinstalling it. Not exactly sure where this is coming from and this is quite a dirty patch, but it works.

Is there any known issues with Dissemin and Debian Stretch? Might be interesting to update the box used in Vagrant config to have a more up to date Debian.

Best,

### DIFF
--- a/provisioning/provision.sh
+++ b/provisioning/provision.sh
@@ -1,14 +1,18 @@
 #!/bin/bash
+# Stop execution on first error
+set -e
+# Show commands as they are executed
+set -x
 
 # We update the apt-get cache
 apt-get update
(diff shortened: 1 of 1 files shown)


Phyks:
Yw, thanks

Issue #544, opened 2018-12-19
Publishers page gives a 500


Phyks:
https://dissem.in/publishers gives a 500, it should probably not.

I think we should probably emphasize this page as well, to ease discovery as this is a nice UI on top of Romeo! :)

PR #549, opened 2019-01-08
Fix the 500 error on /publishers

Phyks:
Closes https://github.com/dissemin/dissemin/issues/544.


wetneb:
Thanks! Will review ASAP.

PR #555, opened 2019-01-15
Fix an issue when no affiliations can be found

Phyks:
See https://sentry.io/dissemin/dissemin/issues/744452148/?query=is%3Aunresolved


wetneb:

PR #556, opened 2019-01-15
Return a 404 when DOI or pk is not given in paper view

Phyks:
See https://dissem.in/10.1blabla, this was resulting in a 500 error,
return instead a 404 error.

See as well https://sentry.io/dissemin/dissemin/issues/769971765/.


wetneb:

Issue #512, opened 2018-01-08
False positive with merging heuristics


wetneb:
Two papers are incorrectly merged into one here:
https://dissem.in/p/1957700/early-views-on-binocular-rivalry
We need to investigate and understand what is happening.

nemobis:
There are some extreme cases too, for instance:

* https://dissem.in/p/66582540/bericht-uber-patente (117 DOI links)
* https://dissem.in/p/47544175/extracts-from-the-records-of-the-boston-society-for-medical-improvement (47 DOI links)
* https://dissem.in/p/42904051/short-notices (227 DOI links)

I suppose it's not trivial to keep the merging/clustering useful while requiring stricter criteria (e.g. requiring a perfect match for all of title, author and date). So I would suggest to just delete/hide for now all the records with more than 10 (50? 100?) links.

How long does it take to redo an import and merge from scratch?

nemobis:
This article from 1857 is listed as being from 2014 in the Dissemin header, presumably as leftover from ResearchGate (where 2014 is used as date):
https://dissem.in/p/41405635/recherches-sur-quelques-plantes-alimentaires-de-tahiti-iles-de-la-societe

wetneb:
The first example is solved by #691. The rest are probably genuine false positives of the fingerprinting heuristic.

nemobis:
Another case: https://dissem.in/p/70990242/rheumatic-heart-disease, [found by Graham87](https://en.wikipedia.org/?diff=891310602). This may be made worse by PubMed scanned titles which don't have much usable metadata. The solution is the same though: there's no way 58 URLs can be legitimately merged, so if we stop much earlier than that any damage will be strongly limited.

PR #557, opened 2019-01-17
Reduce import/merging errors

nemobis:
Some papers were skipped or overzealous clusters were created.

See issue #512

### DIFF
--- a/backend/crossref.py
+++ b/backend/crossref.py
@@ -405,7 +405,8 @@ class CrossRefAPI(object):
         if metadata is None or type(metadata) != dict:
             raise ValueError('Invalid metadata format, expecting a dict')
         if not metadata.get('author'):
-            raise ValueError('No author provided')
+            # BareName has "last" as mandatory field
+            metadata['author'] = {'family': 'N/A'}
 
         if not metadata.get('title'):
             raise ValueError('No title')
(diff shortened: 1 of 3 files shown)


backend/crossref.py
```diff
@@ -405,7 +405,8 @@ def save_doi_metadata(self, metadata, extra_orcids=None):
         if metadata is None or type(metadata) != dict:
             raise ValueError('Invalid metadata format, expecting a dict')
         if not metadata.get('author'):
-            raise ValueError('No author provided')
+            # BareName has "last" as mandatory field
+            metadata['author'] = {'family': 'N/A'}
```
wetneb: This code is never going to be executed because it is just after an exception is raised.

wetneb:
This would need to be motivated by a careful analysis of the existing conflicts, I think (taking dissemin papers with a lot of oai records and understanding how we got to this situation).

wetneb:
To be more precise: first, thanks for the PR, it's a very important issue.
Second, we should have test cases for this (many test cases demonstrating which papers get merged and do not get merged). This will be easier once the tests are fixed.
Third, we should think about the workflow of un-merging papers. If a paper has not been added to any profile and not uploaded we can just delete it and re-import all its records independently, but otherwise it seems a bit complicated to me.

nemobis:
Antonin Delpeuch, 03/02/19 16:27:
> @nemobis <https://github.com/nemobis> Travis is functional again, so the 
> build log should help you find bugs in the PR :)

Thank you, will give a look by the weekend.

Issue #330, opened 2016-11-26
reason: Undefined


cpitclaudel:
Both messages in this screenshot look fishy:

![image](https://cloud.githubusercontent.com/assets/8181630/20641539/44f1da30-b3c8-11e6-9ff8-90b4cb013127.png)

My ORCID profile does have papers, and reason: undefined is not very informative :)

wetneb:
thanks, it's quite ugly indeed. By the way, are we talking about https://orcid.org/0000-0002-1900-3901? If I'm not mistaken, this profile does not have any name (or at least no public name). I think the profile should be rejected in the first place (#327).

cpitclaudel:
Yeah. Wouldn't it be helpful if orcid.org had a clear way to make a profile public?

cpitclaudel:
Should be fixed now.  Don't ask me why ORCID's privacy settings widget is operated by a third party that uses a tracking cookie.

wetneb:
thanks! yeah their interface is quite confusing.

wetneb:
Note: there is still a problem with papers not being imported from the ORCID profile for some reason.

Phyks:
Now that #327 is merged, the first root cause of this issue should be fixed, I guess. Looking at the Dissem.in profile matching the ORCID ID in https://github.com/dissemin/dissemin/issues/330#issuecomment-263072163, it seems all the (public) papers from ORCID are imported in Dissemin.

Is there still an issue here? Or should it be closed?

wetneb:
The only outstanding issue here is that the messages should be readable and their wording should be improved, I think.

Issue #327, opened 2016-11-26
Refuse ORCID profiles with no public name


wetneb:
Because they're annoying to display and mess up our ORCID import tools.

wetneb:
Live example: https://orcid.org/0000-0002-8735-3014
Sandbox example:  http://sandbox.orcid.org/0000-0002-6091-2701

PR #565, opened 2019-01-25
Reject ORCID accounts without names

### LINKED ISSUE #502: Remove obsolete page that asks for name
Apparently https://github.com/dissemin/dissemin/blob/master/papers/templates/papers/createProfile.html has been displayed when logging in for the first time with 0000-0002-4471-4000. We should find out why and remove this obsolete page.

Phyks:
The previously existing code fixing #327 did reject ORCID profiles with
private name in the "Researcher" view but was not rejecting such
profiles at sign up. This should now be fixed.

Additionally, this removes the obsolete "createProfile" page asking for
a name when no name is provided.

Should fix #502 and probably #330.

This is still quite a proof of concept and error messages displayed to the user should be polished further. I was thinking about an error page (similar to 404 page) stating that Dissemin requires access to the name on the ORCID profile ("trusted parties" and "public" settings from ORCID should be fine I think. There could probably be a link to https://support.orcid.org/hc/en-us/articles/360006897614-Visibility-settings.

I don't have much experience with Django templates and HTTP responses, if anyone is willing to polish the error messages, feel free to continue :) Otherwise, I can have it a try.

Best,


papers/views.py
```diff
@@ -85,7 +90,7 @@ def fetch_on_orcid_login(sender, **kwargs):
     else:
         r.fetch_everything_if_outdated()
 
-post_social_login.connect(fetch_on_orcid_login)
+pre_social_login.connect(fetch_on_orcid_login)
```
Phyks: We need `pre_social_login` and not `post_social_login` here to prevent any account creation if we don't have a name.

wetneb:
Closing and reopening to try if it is enough to update Travis.

PR #582, opened 2019-02-04
Fix a SEntry reported error

Phyks:
https://sentry.io/dissemin/dissemin/issues/871750601/?query=is%3Aunresolved

Note: Paper was https://dissem.in/p/2761246/semaphorin-3a-is-an-endogenous-angiogenesis-inhibitor-that-blocks-tumor-growth-and-normalizes-tumor-vasculature-in-transgenic-mouse-models. Username is at the end of sentry report.

### DIFF
--- a/papers/models.py
+++ b/papers/models.py
@@ -978,9 +978,10 @@ class Paper(models.Model, BarePaper):
             user_researcher = Researcher.objects.get(user=user)
         except Researcher.DoesNotExist:
             user_researcher = None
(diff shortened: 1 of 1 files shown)


wetneb:
Looking at sentry it is curious that this user is not associated with any user - I would like to investigate why before merging this.

wetneb:
Actually it can be the case for some users (for instance admin users not created via orcid). So the fix is fine.

PR #584, opened 2019-02-12
Fix import interfaces

wetneb:
These are temporary fixes to resume the imports from data sources. While the imports are catching up, I will then restructure this, removing all references to proaixy.


backend/crossref.py
```diff
@@ -535,25 +537,49 @@ def fetch_all_records(self, filters=None,cursor="*"):
...
+                filters={'from-update-date':last_updated.date().isoformat(),
+                         'until-update-date':until_date.isoformat()},
+                    cursor=starting_cursor)
+            for record in to_update:
+                try:
+                    bare_paper = self.save_doi_metadata(record)
+                    p = Paper.from_bare(bare_paper)
+                    p.update_index()
+                except ValueError as e:
+                    print((record.get('DOI') or 'unknown DOI') + ': '+unicode(e))
```
Phyks: Should probably be `str` rather than `unicode`, no?

Thinking about a possible migration to Python 3, `unicode` will not be valid. `str` will be the same as `unicode` in Python 3. In Python 2, not sure we need a `unicode` here, but if this is the case, this can be handled with Python-future, see http://python-future.org/compatible_idioms.html#unicode.

backend/oadoi.py
```diff
@@ -72,32 +65,32 @@ def create_oairecord(self, record):
...
-            splash_url=url,
-            pdf_url=record['url'])
-        try:
-            paper.add_oairecord(record)
-            paper.update_availability()
-            # TODO re-enable this
-            #paper.update_index()
-        except (DataError, ValueError):
-            print('Record does not fit in the DB')
+        for oa_location in record.get('oa_locations') or []:
```
Phyks: `.get('oa_locations', [])` will do the same.

wetneb: It's not exactly equivalent: `{'oa_locations':None}.get('oa_locations', [])` returns `None`, which is not iterable, whereas  `{'oa_locations':None}.get('oa_locations') or []` returns `[]`, which is iterable. So I generally try to stick to the latter.

backend/crossref.py
wetneb: I didn't know about `from builtins import str`, it sounds great (I thought `six` was needed for this sort of thing).

backend/oadoi.py
Phyks: Indeed, nice tip :)

backend/crossref.py
Phyks: You need the `future` lib to be able to use `builtins` (`pip install future`). Alternative solution is to manage it by hand, such as https://git.weboob.org/weboob/weboob/blob/master/weboob/tools/compat.py#L33-36.

wetneb: I'll take note of this for projects that need to be compatible with both Python 2 and 3. I think I will leave this as it is in our case, as our migration to Python 3 will not be retro-compatible with Python 2 (because we want to migrate to Django 2). So this use of `unicode` should be translated by `2to3` just like the hundreds of others :)

PR #586, opened 2019-02-17
Migrate to Python 3

### LINKED ISSUE #572: Migrate to Python 3
Once #569, #570 and #571 are solved we should be able to migrate to Python 3.

wetneb:
The only problematic dependency is Celery, which is not compatible with Python 3.7 yet, but apparently a release is coming soon (https://github.com/celery/celery/issues/5180), so we should quickly be able to replace the pinned commit with an official version.

Closes #572.


wetneb:
Nevermind, I went for Python 3.6 which is compatible with the latest Celery. We can also add other Python 3 versions to the Travis config if we want - it is probably going to slow builds down a bit though, I think.

wetneb:
Note: to test this PR, make sure you create a fresh virtualenv with Python 3 (<3.7) and remove all your *.pyc files compiled from dissemin's source files.

PR #598, opened 2019-02-18
Document how to selfhost Mathjax

Phyks:
Also include a polyfill for es6-promise, required by the fetch polyfill on older browsers.


requirements.txt
```diff
@@ -22,6 +22,7 @@ django-jsonview
 django-leaflet
 django-memoize
 django-redis-cache
+django-settings-export
```
Phyks: I used this to access some settings from within the template. Not sure if there is a better way?

Phyks:
I'll probably let you do the deployment part. This should be done by

```
cd /home/dissemin/www/static
wget https://github.com/mathjax/MathJax/archive/master.zip
unzip master.zip
```

and setting `MATHJAX_SELFHOST_URL = 'https://dissem.in/static/MathJax/MathJax.js?config=TeX-AMS-MML_HTMLorMML'` for instance.

PR #610, opened 2019-02-19
[requires.io] dependency update on master branch

Phyks:


Phyks:
@wetneb one test is failing with this PR. This is a bit weird, it does not seem to be due to the updates though, at first sight.

`html5lib` should not have any breaking changes https://github.com/html5lib/html5lib-python/blob/master/CHANGES.rst#101.

`pytest` should not either https://github.com/pytest-dev/pytest/blob/master/CHANGELOG.rst#pytest-430-2019-02-16

Is there a specific reason for pinning these to these specific versions (`pytest` in particular) in the `requirements.txt` file? If not, I'll have a deeper look and try to make this one mergeable and implement frozen requirements for the future.

wetneb:
I pinned `pytest` for compatibility with its django plugin I think, but it could also have been for a coverage plugin (`pytest-cov`) that I didn't end up using. But I suspect the failure could be linked to that (different behaviour of fixtures and test set ups?).

For htmllib I can't remember, that was years ago - it's probably safe to upgrade (it's only used for the tests anyway, so as long as the current tests pass it's fairly safe).

Phyks:
I checked and merging the first commit, updating `html5lib` is definitely fine.

The tests are failing due to `pytest` update, I'll investigate this.

Phyks:
@wetneb Ok, I managed to reduce a lot the scope of the issue. I would probably need a second look for fixing this though :/

Minimal command to reproduce is:

```
python -m pytest -x deposit/osf/tests.py papers/tests/test_models.py -k 'test_create_by_identifier_valid_paper or test_get_form_initial_data'
```

And this happens for Pytest >= 4.2.0, so it is due to a change in [4.2.0](https://github.com/pytest-dev/pytest/blob/master/CHANGELOG.rst#pytest-420-2019-01-30).

I traced back the issue to be coming from [this line in `ProtocolTest`](https://github.com/dissemin/dissemin/blob/master/deposit/tests/test_protocol.py#L48). Removing the `@pytest.mark.usefixtures` removes the issue.

Given the changelog, I guess this is linked to https://github.com/pytest-dev/pytest/issues/517 / https://github.com/pytest-dev/pytest/issues/3094, but I'm a bit stuck at this point :/

Issue #597, opened 2019-02-18
Add monitoring for Celery


wetneb:
We should have a way to be notified when Celery unexpectedly stops in production as it stalls all asynchronous tasks. This would happen (for instance) when we run out of disk space.

More generally we might want to improve the way we deploy and update celery (kubernetes?).

wetneb:
For now, here is how deployment is done:
- dissemin itself runs as a WSGI script loaded by Apache
- celery is launched in a screen with `./celery.sh` (as `dissemin` user)
- elasticsearch is launched in a screen with `elasticsearch -d` (as `dissemin` user)

Phyks:
Ok. I guess writing systemd units for celery and elasticsearch could help here. There would be auto-restart for free basically, and eventually some notifications.

Phyks:
Concerning elasticsearch, I think we should probably install it through the `deb` packages, https://www.elastic.co/guide/en/elasticsearch/reference/current/deb.html.

This would probably ease maintenance and upgrade. Additionnally, it comes with a SysV/Systemd unit https://www.elastic.co/guide/en/elasticsearch/reference/current/deb.html#deb-running-systemd. This will provide autostart.

Autorestart is not provided by the unit, but this can be easily added on top.


Writing a unit for celery should not be very difficult either.

beckstefan:
celery is slightly tricky, since their docs do not say anything about celery _beat_, which can indeed be run as daemon, but it lacks a proper stop command.

For celery _beat_ systemd unit:
https://github.com/celery/celery/issues/4304

wetneb:
Done as part of the migration to the new servers.

Issue #611, opened 2019-02-20
Depositing is broken


Phyks:
Deposit page is currently broken.

There is a first error coming from django-crispy-forms

> new_render() got an unexpected keyword argument 'renderer'

This is https://github.com/django-crispy-forms/django-crispy-forms/issues/852, which is actually due to https://github.com/yourlabs/django-autocomplete-light/issues/1026.

It should have been fixed upstream but it seems it somehow is not :/ Downgrading Django to 2.0.8 is a workaround.

But then, we get another issue

> '__proxy__' object is not callable

Which is apparently coming from https://github.com/django-crispy-forms/django-crispy-forms/blob/master/crispy_forms/templates/bootstrap3/field.html#L28. Not sure what is happening there :/

PR #612, opened 2019-02-20
More documentation

beckstefan:
With the migration to Python3 and Django2 there were some documentation updates necessary.

Also concerning  #597 there is now a documentation for celery and celerybeat running as daemon.

### DIFF
--- a/doc/sphinx/deploying.rst
+++ b/doc/sphinx/deploying.rst
@@ -21,7 +21,7 @@ Production settings
 
 As any Django website, Dissemin can be served by various web servers.
 These settings are not specific to dissemin itself so you should refer
-to `the relevant Django documentation <https://docs.djangoproject.com/en/1.11/howto/deployment/>`_.
+to `the relevant Django documentation <https://docs.djangoproject.com/en/2.2/howto/deployment/>`_.
 
 No matter what web server you use,
 you need to run ``python manage.py collectstatic`` to copy the static files from
@@ -59,7 +59,7 @@ to load MathJax from. In the example below, this would be
 Apache with WSGI
 ----------------
 
-Here is a sample VirtualHost, assuming that the root of the Dissemin source code is at ``/home/dissemin``.::
+Here is a sample VirtualHost, assuming that the root of the Dissemin source code is at ``/home/dissemin`` and you use ``python3.6``.::
 
     <VirtualHost *:80>
             ServerAdmin webmaster@localhost
@@ -93,7 +93,7 @@ Here is a sample VirtualHost, assuming that the root of the Dissemin source code
             # Path to the WSGI application for the website
             WSGIScriptAlias / /home/dissemin/dissemin/wsgi.py
             # Python path for the application
-            WSGIDaemonProcess dissemin.myuni.edu python-path=/home/dissemin:/home/dissemin/.virtualenv/lib/python2.7/site-packages
+            WSGIDaemonProcess dissemin.myuni.edu python-path=/home/dissemin:/home/dissemin/.virtualenv/lib/python3.6/site-packages
 
             WSGIProcessGroup dissemin.myuni.edu
 
--- a/doc/sphinx/install.rst
+++ b/doc/sphinx/install.rst
@@ -55,12 +55,12 @@ Installation instructions for the web frontend
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 First, install the following dependencies (debian packages)
-``postgresql postgresql-server-dev-all postgresql-client python-virtualenv build-essential libxml2-dev libxslt1-dev python-dev gettext libjpeg-dev libffi-dev``
+``postgresql postgresql-server-dev-all postgresql-client python3-venv build-essential libxml2-dev libxslt1-dev python3-dev gettext libjpeg-dev libffi-dev``
 
 Then, build a virtual environment to isolate all the python
 dependencies::
 
-   virtualenv .virtualenv
+   python3 -m venv .virtualenv
    source .virtualenv/bin/activate
    pip install --upgrade setuptools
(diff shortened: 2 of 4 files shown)


doc/sphinx/deploying.rst
```diff
@@ -21,7 +21,7 @@ Production settings
 
 As any Django website, Dissemin can be served by various web servers.
 These settings are not specific to dissemin itself so you should refer
-to `the relevant Django documentation <https://docs.djangoproject.com/en/1.11/howto/deployment/>`_.
+to `the relevant Django documentation <https://docs.djangoproject.com/en/2.2/howto/deployment/>`_.
```
Phyks: Not sure about the relevant Django documentation here, with https://github.com/dissemin/dissemin/issues/611 in mind. So far, dissem.in is in 2.0.8 I think, but the issue is not solved yet :/

Same below when linking to Django doc.

doc/sphinx/install.rst
```diff
@@ -55,12 +55,12 @@ Installation instructions for the web frontend
...
 
 First, install the following dependencies (debian packages)
-``postgresql postgresql-server-dev-all postgresql-client python-virtualenv build-essential libxml2-dev libxslt1-dev python-dev gettext libjpeg-dev libffi-dev``
+``postgresql postgresql-server-dev-all postgresql-client python3-venv build-essential libxml2-dev libxslt1-dev python3-dev gettext libjpeg-dev libffi-dev``
 
 Then, build a virtual environment to isolate all the python
 dependencies::
 
-   virtualenv .virtualenv
+   python3 -m venv .virtualenv
```
Phyks: I updated the provisioning script recently and am using the `python3-virtualenv` package (https://github.com/dissemin/dissemin/blob/master/provisioning/provision.sh#L27). Do you know what is the difference with `python3-venv`? Thanks!

doc/sphinx/install.rst
```diff
@@ -238,6 +238,69 @@ To run the backend (still in the virtualenv)::
...
+
+    [Service]
+    Type=forking
+    User=dissemin
+    Group=dissemin
+    EnvironmentFile=-/etc/default/celery
+    WorkingDirectory=/path/to/dissemin/
+    ExecStart=/bin/sh -c '${CELERY_BIN} multi start ${CELERYD_NODES} -A ${CELERY_APP} --pidfile=${CELERYD_PID_FILE} --logfile=${CELERYD_LOG_FILE} --loglevel=${CELERYD_LOG_LEVEL} ${CELERYD_OPTS}'
+    ExecStop=/bin/sh -c '${CELERY_BIN} multi stopwait ${CELERYD_NODES} --pidfile=${CELERYD_PID_FILE}'
+    ExecReload=/bin/sh -c '${CELERY_BIN} multi restart ${CELERYD_NODES} -A ${CELERY_APP} --pidfile=${CELERYD_PID_FILE} --logfile=${CELERYD_LOG_FILE} --loglevel=${CELERYD_LOG_LEVEL} ${CELERYD_OPTS}'
```
Phyks: With #597 in mind, don't we want something like `Restart=always` in this systemd unit?

Same question below for `celeryd`.

doc/sphinx/install.rst
```diff
@@ -238,6 +238,69 @@ To run the backend (still in the virtualenv)::
...
+    Description=Celery Service
+    After=network.target
+
+    [Service]
+    Type=forking
+    User=dissemin
+    Group=dissemin
+    EnvironmentFile=-/etc/default/celery
+    WorkingDirectory=/path/to/dissemin/
+    ExecStart=/bin/sh -c '${CELERY_BIN} multi start ${CELERYD_NODES} -A ${CELERY_APP} --pidfile=${CELERYD_PID_FILE} --logfile=${CELERYD_LOG_FILE} --loglevel=${CELERYD_LOG_LEVEL} ${CELERYD_OPTS}'
```
Phyks: Could you add a comment here that you need to call `/bin/sh -c` to be able to actually process the `CELERY_BIN` environment variable, please? (same below for `celeryd`).

I was a bit puzzled at first by this syntax :)

doc/sphinx/install.rst
```diff
@@ -238,6 +238,69 @@ To run the backend (still in the virtualenv)::
...
+
+    [Install]
+    WantedBy=multi-user.target
+
+To make systemd create the necessary directories with permissions put the follwing into ``/etc/tmpfiles.d/celery.conf``::
+
+    d /var/run/celery 0755 dissemin dissemin
+    d /var/log/celery 0755 dissemin dissemin
+
+After that run ``systemctl daemon-reload`` to reload systemd service files and you are ready to use ``celery`` and ``celerybeat`` with systemd.
```
Phyks: Maybe we should mention the `systemctl enable` and `systemctl start/stop` briefly (this is not supposed to be a Systemd how to, so no details are needed, just the basic commands to enable and start the services)?

Phyks:
Haha, @wetneb merged this one while I was sending questions. Nothing blocking a merge anyway, thanks a lot for the documentation @beckstefan!

I still posted my comments, these are rather questions regarding open issues and minor details.

wetneb:
Sorry @Phyks - I should have waited, there was no hurry. (I thought this would be better viewed on ReadTheDocs.) I will guard my enthusiasm next time!

doc/sphinx/deploying.rst
wetneb: I will investigate and fix this shortly

doc/sphinx/install.rst
wetneb: I think `python3 -m venv` is recommended over the `virtualenv` command as it runs in python directly. I think the latter is even sort of deprecated.

doc/sphinx/deploying.rst
wetneb: Fixed, the docs are still current.

PR #616, opened 2019-02-21
More details for installation of celery and celerybeat as daemons

beckstefan:
@Phyks recommended to add more explanations regarding the systemd units for celery and celerybeat and to include auto restart.


Phyks:
Thanks!

PR #620, opened 2019-02-21
Fix Sentry issue

Phyks:
https://sentry.io/organizations/dissemin/issues/894371243/?project=189330&query=is%3Aunresolved&statsPeriod=14d&utc=true.

Should be fixed I think, according to https://github.com/dissemin/dissemin/blob/master/papers/urls.py#L33-L34.

### DIFF
--- a/papers/views.py
+++ b/papers/views.py
@@ -321,9 +321,9 @@ class JournalPapersView(PublisherPapersView):
 
 
 @user_passes_test(is_authenticated)
-def refetchResearcher(request, pk):
+def refetch_researcher(request, pk):
     researcher = get_object_or_404(Researcher, pk=pk)
-    if researcher.user != request.user and not request.user.is_staff:
+    if researcher.user != request.user and not (request.user.is_staff or request.user.is_superuser):
         return HttpResponseForbidden("Not authorized to update papers for this researcher.")
     from backend.tasks import fetch_everything_for_researcher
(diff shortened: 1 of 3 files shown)


papers/views.py
```diff
@@ -328,7 +328,7 @@ def refetchResearcher(request, pk):
     from backend.tasks import fetch_everything_for_researcher
     fetch_everything_for_researcher.delay(pk=pk)
 
-    return redirect(reverse('researcher', researcher=pk))
+    return redirect(reverse('researcher', slug=researcher.slug))
```
wetneb: I think the pk should be required too - we cannot identify a researcher by its slug only. I assume that this part of the code is not tested yet.

Phyks: Indeed, I'm rebuilding the VM at the moment. We need both researcher and slug I think.

wetneb: and we need it to be wrapped in an `args={...}` keyword too (hence the error)

wetneb: anyway it would be good to confirm that with a unit test (I can add it)

Phyks: Tested in my VM and fixed. There was actually already a test, except it is only checking the case when the researcher is not logged in. Not sure how to fix this test :/ https://github.com/dissemin/dissemin/pull/620/files#diff-2e713353c3d10b8fe4b78e8759e15dc0R116

Phyks:
Thanks for the unittest!

PR #623, opened 2019-02-22
Romeo improvements

wetneb:
Various tweaks to the RoMEO interface after running the updates in productions. Luckily nothing too serious.


publishers/models.py
```diff
@@ -112,10 +114,12 @@ class Meta:
     def find(cls, publisher_name):
         """
         Lookup a publisher by name. Return None if could not be found.
+        This restricts the search to default policies (those w
```
Phyks: Truncated line :)

publishers/romeo.py
```diff
@@ -79,7 +77,7 @@ def perform_romeo_query(self, search_terms):
         # Parse it
         try:
             parser = ET.XMLParser(encoding='ISO-8859-1')
-            root = ET.parse(BytesIO(response), parser)
+            root = ET.parse(BytesIO(req.content), parser)
```
Phyks: What about something like `root = ET.fromstring(req.text)`? This would let `requests` do the encoding handling on its own, maybe safer?

wetneb: It looks like it works indeed! I think this approach was inherited from the days of urlopen…

publishers/models.py
wetneb: Good catch!

publishers/romeo.py
wetneb: Actually no:
`E   ValueError: Unicode strings with encoding declaration are not supported. Please use bytes input or XML fragments without declaration.`

wetneb: So I think we can leave it like that - SHERPA declares its encoding in the XML so we can leave the encoding handling to lxml.

Phyks: Sure

Issue #514, opened 2018-01-15
Search by ORCID


a3nm:
When searching by author, it should be possible to search with an ORCID ID, and get all papers by the corresponding ORCID ID.

Phyks:
So, I think I could handle this:

1. On the front page, in the search field "any author", if the query matches an ORCID ID, we could simply redirect to the `/<ORCID_ID>` page.
2. On the search page, there would be an extra "by ORCID" field to search by ORCID (multiple values possible, split on space). We could probably use some frontend item such as https://bootstrap-tagsinput.github.io/bootstrap-tagsinput/examples/.

I just have to look up how search is working backend-side to add a filter on ORCID values.

What do you think about this plan?

wetneb:
Item 1 sounds like a no-brainer indeed!
For item 2, we should be careful of not overloading this search form too much - I think it is already quite confusing. Some of these fields could be moved to an "advanced search" page.

What is the motivation for searching for multiple ORCID ids? Do you have a particular use case in mind?

wetneb:
Also, if we let people input orcid ids in the search form on the main page, then I think by consistency they should also be allowed in the authors field of the form on https://dissem.in/search/ - so I would lean towards just reusing this field and not introducing a new one.

Phyks:
> What is the motivation for searching for multiple ORCID ids? Do you have a particular use case in mind?

I was just having the symmetry in mind "by authorS" (multiple authors possible), so by ORCIDs.

> Also, if we let people input orcid ids in the search form on the main page, then I think by consistency they should also be allowed in the authors field of the form on https://dissem.in/search/ - so I would lean towards just reusing this field and not introducing a new one.

So whatever word (split on whitespace) matches an ORCID ID regex, should be considered as an ORCID ID and added to the search filter as such. The rest of the words will be passed down to the authors such as currently done.

wetneb:
Note: currently, this will only work for ORCID profiles that we have already fetched. We have plans to fetch all of them anyway.

PR #626, opened 2019-02-25
Search by ORCID, #514

Phyks:
This is a work in progress for solving #514.

Nothing is broken (yet :p) but the further filtering on ORCIDs is not working.

Estimated time: about one - two hours.

### DIFF
--- a/papers/views.py
+++ b/papers/views.py
@@ -124,6 +124,7 @@ class PaperSearchView(SearchView):
             request.GET.pop('visible', None)
             request.GET.pop('availability', None)
             request.GET.pop('oa_status', None)
+
         return super(PaperSearchView, self).get(request, *args, **kwargs)
 
     def get_context_data(self, **kwargs):
@@ -160,7 +161,8 @@ class PaperSearchView(SearchView):
         args = super(PaperSearchView, self).get_form_kwargs()
 
         if 'data' not in args:
-            args['data'] = {self.search_field:''}
+            args['data'] = {self.search_field: ''}
+
         return args
 
     def render_to_response(self, context, **kwargs):
--- a/papers/forms.py
+++ b/papers/forms.py
@@ -141,6 +141,21 @@ class PaperForm(SearchForm):
         # Items with no whitespace of prefixed with 'last:' are considered as
         # last names; others are full names.
         for name in self.cleaned_data['authors'].split(','):
+            name = name.strip()
+
+            # If part of this author name matches ORCID identifiers, consider
+            # these as orcid ids and do the filtering
+            orcid_ids = [x for x in name.split(' ') if validate_orcid(x)]
+            for orcid_id in orcid_ids:
+                try:
+                    researcher = Researcher.objects.get(orcid=orcid_id)
+                    self.filter(researchers=researcher.id)
+                except Researcher.DoesNotExist:
+                    pass
+                continue
+            # Rebuild a full name excluding the ORCID id terms
+            name = ' '.join([x for x in name.split(' ') if x not in orcid_ids])
+
             name = remove_diacritics(name.strip())
 
             if name.startswith('last:'):


papers/views.py
```diff
@@ -124,6 +124,28 @@ def get(self, request, *args, **kwargs):
...
+            orcids = [
+                item
+                for item in query['authors'].split()
+                if validate_orcid(item)
+            ]
+            for orcid in orcids:
+                try:
+                    researcher = Researcher.objects.get(orcid=orcid)
+                    researchers.append(researcher.id)
+                except MetadaSourceException:
```
wetneb: I would expect this exception to be `Researcher.DoesNotExist` instead.

papers/views.py
```diff
@@ -124,6 +124,28 @@ def get(self, request, *args, **kwargs):
...
+                    pass
+            authors = [
+                item
+                for item in query['authors'].split()
+                if item not in orcids
+            ]
+            if researchers:
+                self.queryset = self.queryset.filter(researchers=researchers)
```
wetneb: I am not sure about the syntax for SearchQuerySets but for Django's standard query sets, filtering on a property with a list of possible values, you need to use `filter(researchers__in=researchers)`.

papers/views.py
Phyks: Oh, OK. This is copy pasted from https://github.com/dissemin/dissemin/blob/master/papers/views.py#L212-L219 actually.

papers/views.py
Phyks: I probably have to do the same indeed, thanks for the tip! Even if `researchers` is not a list however, it does not work. I probably have a logic error here, I'll check this.

papers/views.py
wetneb: This exception is thrown by `researcher.init_from_orcid()` which you have removed. Maybe it would make sense to refactor the code you copied into a function that can be reused for both views?

wetneb: I am not sure though - the views could be doing different things. The question is: do we want to trigger fetches from ORCID in Celery if the ORCID id supplied is not in dissemin yet? We do that for the https://dissem.in/<orcid> view, but not sure if we want to put that in the search interface itself…

a3nm: Unrelated but there's a typo: "MetadaSourceException" should be "MetadataSourceException". Note that this typo is not in https://github.com/dissemin/dissemin/blob/master/papers/views.py#L212-L219

Phyks: Fixed and tested.

Phyks:
So, in principle this should be working. From the Django-Haystack doc and other Dissemin search views, it seems I'm not doing anything wrong.

From the Python / Django shell, manually creating `SearchQuerySet().models(Paper).filter(researchers__in=[SOME_RESEARCHER_ID])` does work.

Still, I'm apparently missing something and the results page is always empty when querying with a unique ORCID ID (same search with the researcher name works fine). :/

Phyks:
Small update on this.

https://github.com/dissemin/dissemin/pull/626/commits/8bb497cb97bd2e2dba79a343989ff290c3c8a34a is a working version. Except I'm actually messing the search query, so the "authors" field in the form in the search sidebar is no longer working (not showing the ORCID terms).

I came across https://github.com/dissemin/dissemin/blob/master/papers/forms.py#L143 which actually definitely looks like the right place to do the tricks. This is implemented in https://github.com/dissemin/dissemin/pull/626/commits/f6c0544ebee76b7387073f53bdc0a35eb24ca263, but it does not work so far. Not sure why, it seems the query issued to ElasticSearch is the right one.

In ElasticSearch logging, I can see

> [2019-02-28 13:20:53,504][DEBUG][index.search.slowlog.query] [Kleinstocks] [dissemin][4] took[8.2ms], took_millis[8], types[modelresult], stats[], search_type[QUERY_THEN_FETCH], total_shards[5], source[{"aggs":{"status":{"terms":{"field":"combined_status_exact"}}},"from":0,"sort":[{"pubdate":{"order":"desc"}}],"query":{"filtered":{"filter":{"terms":{"django_ct":["papers.paper"]}},"query":{"query_string":{"fuzzy_min_sim":0.5,"default_operator":"AND","analyze_wildcard":true,"default_field":"text","fuzzy_max_expansions":50,"auto_generate_phrase_queries":true,"query":"(visible:(True) AND researcher:(1))"}}}}}], extra_source[{"_source":{"includes":[],"excludes":[]}}],

which seems to be fine to return papers from the researcher with id 1.

I cannot debug it much further right now but will have a look at it soon to finish this work.

Phyks:
I got a bit of time to finish working on this, the latest pushed version should be working fine.

Tested with:

* http://localhost:8080/search/?authors=0000-0002-9658-1473%2C+gunjal
* http://localhost:8080/search/?authors=0000-0002-9658-1473+gunjal
* http://localhost:8080/search/?authors=0000-0002-9658-1473
* http://localhost:8080/search/?authors=gunjal

Please let me know if you find any bug :)

Issue #630, opened 2019-02-28
Use units for the number of papers in front page


Phyks:
Hi,

On the front page, we display the total number of papers in each category. See

![capture d ecran 2019-02-28 a 12 15 52](https://user-images.githubusercontent.com/3856586/53562805-aac67280-3b52-11e9-90de-bd80f1d00d1d.png)

I think we should use units (display "9k", "4M", etc) rather than the full numbers. We don't care so much about the precise value (unless there was some realtime update and we see the numbers rolling) and the long numbers are difficult to parse so it potentially makes it a bit more difficult for users and newcomers to realize the magnitude of these numbers.

I guess "4.4M papers" would be clearer and more impressive than "4407923". :)

wetneb:
It's probably just a matter of reusing the shortening logic that we already have for the numbers on the diagram.

beckstefan:
I like the concrete numbers more and find them more impressive. Maybe we could use a bootstrap table with category and numbers (numbers right aligned). 

For example we could introduce 4 boxes: Green OA, Open Repositories, Diagram and one for a table.

To make the numbers more readable we could use separators, i.e. 4.407.923, but I don't how we solve translations problems, e.g. in Germany we use commata instead of dots.

wetneb:
Yes separators in these numbers would also be a nice way to make them more readable. Django must certainly have support for localizing that.

a3nm:
No strong opinion here. The exact numbers are essentially illegible (as you can't tell the order of magnitude) but yeah they are somewhat impressive, and in any case the shortened numbers are also on the diagram. I'd have a vague preference for shortening both numbers but it doesn't seem so important to me. :)

Phyks:
Having a readable string with the precise number but separators (localized) is perfectly fine in my opinion :)

Seems Django can actually handle it out of the box https://stackoverflow.com/questions/13335554/number-format-localized-in-django.

wetneb:
Modulo the caveat that charts updated in javascript will not be affected by the change, this was solved in #635, so I'll close this.

Phyks:
Indeed, I totally missed that JS could update the numbers. I'll look for a fix for this, should not be much harder :)

Phyks:
Should be fixed now :)

PR #635, opened 2019-03-14
Use thousands separator, very quick fix for #630.

Phyks:
Estimated time: 10 minutes.


wetneb:
Oh wow, that is efficient!

PR #636, opened 2019-03-15
Fix year formatting, no thousands separators

Phyks:
My change from yesterday was actually too simple and was also affecting the years :/

Screenshot_2019-03-15 Papiers - Dissemin

I think the best approach is probably to default to not using thousands separators and explicitly force them for the statistics template. This PR should do it. However, I could not test it so not to be merged without explicitly checking it. I can check it and revise later.


wetneb:
It Works™

PR #640, opened 2019-03-17
Update numbers with thousands separators in JS part

Phyks:
I think the only remaining point to solve #630 was the missing thousands separators for the total number of papers field in the search view. This PR should fix it.

Estimated Time: 45 mins.


wetneb:
@Phyks I've deployed it but it does not seem to fix the issue completely?

Phyks:
Hmmm, indeed in French locales separators used are commas. That's weird :/ I'll investigate this.

Phyks:
@wetneb this should be ready to merge and working now. I also fixed an error in JS console and should have enabled Sentry for JS code.

Phyks:
Note: I broke the translations by moving the sentence from one file to another apparently. Not sure how I should restore this…

dissemin/settings/common.py
```diff
@@ -155,6 +155,7 @@
     'django_countries',
     'leaflet',
     'djgeojson',
+    'raven.contrib.django.raven_compat',
```
wetneb: I don't think this should be enabled in `common.py`, as it would also affect dev instances. Also this module is deprecated according to Sentry's docs:
https://docs.sentry.io/clients/python/integrations/django/
We currently use the newer version described here:
https://docs.sentry.io/platforms/python/django/

Phyks: Should be done in https://github.com/dissemin/dissemin/pull/640/commits/53c58db5af1bb2de4b76d3bf4417beb981f67d11. Just be careful that a small migration should probably be done in prod to put the Sentry config in the right place.

papers/templates/papers/search.html
```diff
@@ -201,7 +202,13 @@
...
 <span class="detailsTitle">{% trans "Search" %}</span>
 <div class="detailsContent">
-<p id="nbPapersFound">{{ nb_results }}</p>
+    <p id="nbPapersFound">
+        {% blocktrans trimmed with formattedCounter=nb_results|intcomma count counter=nb_results|default:0 %}
+            {{ formattedCounter }} paper found
+            {% plural %}
+            {{ formattedCounter }} papers found
+        {% endblocktrans %}
+    </p>
```
wetneb: It makes sense to do this on the frontend! And the translation system is much better like that. However, this causes issues when dynamically updating the number of results in Javascript: the " papers found" text is not appended to the number. I think we need to use the JS localization system here.

papers/static/papers/refresh.js
```diff
@@ -172,7 +172,11 @@ function init_paper_module (config) {
   function updateData (data) {
     config.paperSearchResultsNode.html(data.listPapers)
     updateStats(data.stats)
-    config.nbPapersFoundNode.text(data.nb_results)
+    var thousandSeparator = get_format('THOUSAND_SEPARATOR')
+    // Formatting with thousand separator is coming from
+    // https://stackoverflow.com/a/2901298
+    var formattedNbResults = data.nb_results.toString().replace(/\B(?=(\d{3})+(?!\d))/g, thousandSeparator)
+    config.nbPapersFoundNode.text(formattedNbResults)
```
wetneb: This successfully adds separators to the total number of papers found, but I think we need to add this in `updateStats` as well, to update the individual counts for each category. So it would probably make sense to package this as a function.
(And a call to the JS localization system will be needed here, to add the "papers found")

wetneb:
(Also: I have temporarily deployed this branch in production)

Phyks:
TODO: When pressing "Enter" in the search form to further refine it, page is reloaded. When just losing focus of the current input, JS magic is ran.

Phyks:
It's not super easy to test in the VM, since I have very few papers loaded, but I guess this should be fixed now :) Thanks for the review!

wetneb:
@Phyks deployed. There seems to be issues still, check the console (or Sentry probably ^^)

Phyks:
Hmm, I cannot trigger any issue in my console (and there is nothing on Sentry) ^^

wetneb:
@Phyks oh sorry, it was a cache issue on my side! Thanks, let's finally merge this!

Issue #516, opened 2018-03-27
"FirstName LastName" should be the same as "LastName FirstName" 


Bisaloo:
I had a colleague complaining they could only find one their own papers on Dissemin today.

Turns out they were searching with "LastName FirstName" instead of "FirstName LastName". I feel like Dissemin could probably be a bit more clever here.

I can provide the name of this person by email if needed. I'd rather not post it on a public bug tracker without their explicit agreement though.

Phyks:
Closed by #642.

PR #642, opened 2019-03-18
Fix issue 516

Phyks:
Would deserve a second look as this is changing the search logic.

This should make "FirstName LastName" the same as "LastName FirstName" and solve #516.

Estimated Time: 15 minutes.

### DIFF
--- a/papers/forms.py
+++ b/papers/forms.py
@@ -28,6 +28,7 @@ from django import forms
 from django.utils.translation import ugettext_lazy as _
 from haystack import inputs
 from haystack.forms import SearchForm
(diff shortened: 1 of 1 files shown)


wetneb:
Looks good to me - let's check that the queries incurred are not massively more expensive than the current ones. On a related note, we still have errors in the elasticsearch logs about queries with incorrect syntaxes, which could be linked to our search syntax for author names.

wetneb:
Merging this now that we have a reliable elasticsearch - we'll see how it goes in production and tweak if performance issues arise.

PR #645, opened 2019-03-19
Fix invalid HTML markup on deposit page

Phyks:


wetneb:
I guess this means our unit tests don't cover this page yet. It could be an easy coverage win to add a test. Our `RenderingTest` makes it easy to run generic checks on pages, including correctness of the HTML code.

Phyks:
The test should have been added now :)

Issue #186, opened 2016-01-23
Add bibtex export [$15]


wetneb:
It could be possible to generate bibtex records from paper pages (and ideally in other formats too).

cpitclaudel:
Connects to #211 (doai.org content negociation) :)

wetneb:
See also issue #219

Phyks:
Which record should be exported? Typically, for [this paper](http://dissem.in/p/8652/autonomization-of-monoidal-categories) should there be an "export to Bibtex" button for any available record or just for the paper itself?

I think this issue could be solved quite easily with code such as https://github.com/Phyks/libbmc/blob/master/libbmc/doi.py#L188-L209 (for records with DOIs), https://github.com/Phyks/libbmc/blob/master/libbmc/repositories/arxiv.py#L276-L306 (for arXiv, and similar codes for any repo-specific exports) and https://github.com/Phyks/libbmc/blob/master/libbmc/bibtex.py#L17-L30 (as a fallback, using the information we have about the paper).

wetneb:
I think it would be simpler to have just one BibTeX record per paper. I think it would be nice to have both the DOI and the link to the preprint in the BibTeX when they are both available. Maybe @a3nm and @p4bl0- have more thoughts about what the perfect BibTeX record should look like?

a3nm:
I guess there could be an option to "export to bibtex", both for individual papers and for pages that list papers (e.g., search results, or list of papers by one author).

I guess we should do like Google Scholar and use LASTNAME2042FIRSTTITLEWORD as a citation key.

It would be nice to include the DOI and URL in a specific field. It would be even nicer if the title of the paper could be clickable to point to the URL or DOI, but doing it by hand (with an `\href`) is ugly. Apparently this can be done by the surrounding document using something [like this](https://tex.stackexchange.com/questions/207004/is-it-possible-to-make-a-reference-clickable-without-showing-its-doi-or-url-fiel) but I've never tried it.

Phyks:
I did a first draft in https://github.com/dissemin/dissemin/pull/647#pullrequestreview-217214655 for a single paper (from the paper details page) export to Bibtex. Please, feel free to comment and review this to fit the need :)

Once single paper export is fine, doing export of a list of results (or profile page) should not be much harder.

Might be worth also letting users export citation to plain text and RIS (used by End Note apparently) as well?

Phyks:
Closing, let us continue the discussion in #666.

PR #647, opened 2019-03-21
First draft of individual paper export, see #186.

Phyks:
This introduces an "Export citation" link in the right hand side bar in the paper detail page

Capture d’écran 2019-03-21 à 13 39 38

Still quite a proof of concept.

### DIFF
--- a/papers/baremodels.py
+++ b/papers/baremodels.py
@@ -613,13 +613,8 @@ class BarePaper(BareObject):
         """
         BibTeX representation of the paper, for citation purposes
         """
-        entrytype = 'misc'
-        for pubtype, bibtex_type in PAPER_TYPE_TO_BIBTEX:
-            if pubtype == self.pubtype:
-                entrytype = bibtex_type
-                break
         entry = {
-            'ENTRYTYPE': entrytype,
+            'ENTRYTYPE': 'misc',
             'ID': (
                 '%s%s' % (
                     self.authors[0].name.last,
@@ -634,6 +629,15 @@ class BarePaper(BareObject):
         }
 
         for publi in self.publications[:3]:
+            try:
+                bibtex_type = next(
+                    b for p, b in PAPER_TYPE_TO_BIBTEX
+                    if p == publi.pubtype
+                )
+                entry['ENTRYTYPE'] = bibtex_type
+            except StopIteration:
+                pass
+
             if publi.volume:
                 entry['volume'] = publi.volume
             if publi.pages:
--- a/papers/baremodels.py
+++ b/papers/baremodels.py
@@ -42,6 +42,7 @@ from django.core.exceptions import ObjectDoesNotExist
 from django.template.defaultfilters import slugify
 from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
+from papers.bibtex import PAPER_TYPE_TO_BIBTEX
 from papers.doi import doi_to_url
 from papers.fingerprint import create_paper_plain_fingerprint
 from papers.utils import datetime_to_date
@@ -612,8 +613,13 @@ class BarePaper(BareObject):
         """
         BibTeX representation of the paper, for citation purposes
         """
+        entrytype = 'misc'
+        for pubtype, bibtex_type in PAPER_TYPE_TO_BIBTEX:
+            if pubtype == self.pubtype:
+                entrytype = bibtex_type
+                break
         entry = {
-            'ENTRYTYPE': 'article',
+            'ENTRYTYPE': entrytype,
             'ID': (
                 '%s%s' % (
                     self.authors[0].name.last,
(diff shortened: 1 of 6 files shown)


papers/baremodels.py
```diff
@@ -601,6 +605,55 @@ def json(self):
             'classification': self.oa_status,
             })
 
+    def bibtex(self):
+        """
+        BibTeX representation of the paper, for citation purposes
+        """
+        entry = {
+            'ENTRYTYPE': 'article',
```
Phyks: Bibtex entry type should probably depend on the `doctype` of the paper. I think we should have a mapping between Dissemin's doctypes and the bibtex equivalents through the ORCID import of bibtex, but I did not yet look into this much.

papers/baremodels.py
```diff
@@ -601,6 +605,55 @@ def json(self):
...
+                )
+            ),
+            'title': self.title,
+            'author': ' and '.join([
+                '%s, %s' % (a.name.last, a.name.first)
+                for a in self.authors
+            ])
+        }
+
+        for publi in self.publications[:3]:
```
Phyks: The choice for the citation export here is to have the URL being a link at which the PDF file can be accessed, if possible and to have a single citation export for an article, based on the DOI and stored values.

Not 100% sure about the `self.publications` logic here, I did not review it carefully since the dev VM does not have enough providers to handle this well I think.

Phyks: I might be missing some BibTeX as well, for this first proof of concept.

wetneb:
Awesome! Happy to merge and deploy this once Travis is green :)
Maybe it would also make sense to add some tests (around the failing test) showing some sample bibtex outputs (even if the output changes in the future)?

Phyks:
Should be done, as well as the correct entrytype in BibTeX. The BibTeX output should be checked though and I should write some tests for this.

If you think the code is fine, we can probably deploy it live, then check with real data that the output is correct and expected and if this is the case I'll complete with tests afterwards.

wetneb:
@Phyks ok!

I think in general we should avoid to deploy other branches than master to production, unless the feature is really hard to test locally. In this case, it should be fairly easy to write basic tests for this: use `Paper.create_by_doi`, generate the bibtex, and do checks on it (for instance parse it back and check the title). But I acknowledge that we need to make it easier to ingest a decent chunk of papers in a small instance, to make local testing easier, so I am just deployed this branch on production.

Phyks:
@wetneb Is there a list of the valid pubtypes somewhere? I was thinking it was https://github.com/dissemin/dissemin/blob/master/papers/baremodels.py#L56-L70, but this does not seem to be the case. When trying some manual fetching of DOIs (for instance `10.1103/physrevapplied.11.024003` or `10.1145/3034786.3056121`), I get quite unexpected `pubtypes` results (from CrossRef OAIRecord) such as "paper-conference" or "article-journal" (not appearing in the previous list).

Anyways, I added unit tests and everything should be working fine. The only negative point so far is that due to my not very good handling of pubtypes, most of papers are exported as `@misc`.

wetneb:
@Phyks note that there is a difference between the type stored at the paper level (`Paper.doctype`) and the one stored at the OaiRecord level (`OaiRecord.pubtype`). I think you should generally rely on the former (because it is supposed to be inferred from the others). That being said I would expect `OaiRecord.pubtype` to be normalized too, I'll investigate.

papers/bibtex.py
```diff
@@ -35,6 +35,22 @@
...
+   ('proceedings', 'proceedings'),
+   ('reference-entry', 'misc'),
+   ('poster', 'misc'),
+   ('report', 'article'),
+   ('thesis', 'phdthesis'),
+   ('dataset', 'misc'),
+   ('preprint', 'article'),
+   ('other', 'misc'),
+]
+
```
wetneb: Is there any particular reason to use a list of pairs rather a dict for `PAPER_TYPE_TO_BIBTEX`? The order of the items does not seem relevant and having a dict would simplify the lookup in the translation, no?

papers/bibtex.py
```diff
@@ -35,6 +35,22 @@
 # Others is a reserved Bibtex keyword
 OTHERS_RE = re.compile(r'( and )?\s*others\.?\s*$', re.IGNORECASE | re.UNICODE)
 
+PAPER_TYPE_TO_BIBTEX = [
+   ('journal-article', 'article'),
+   ('proceedings-article', 'inproceedings'),
+   ('book-chapter', 'book'),
```
wetneb: How about using the `incollection` type for book chapters?

Phyks: :+1: will fix

papers/bibtex.py
Phyks: I just stupidly copy/pasted the `PAPER_CHOICES` constant (which could probably be a dict as well ^^). Will fix.

wetneb: Yeah, for PAPER_CHOICES I was not aware of ordered dicts, which would be the right thing…

Issue #603, opened 2019-02-18
Use proper logging instead of print statements


wetneb:
Because this is quite embarassing!

https://docs.djangoproject.com/en/2.1/topics/logging/

wetneb:
Fixed by #648.

PR #648, opened 2019-03-21
WIP: Setting up logging

beckstefan:
Tackling Issue #603. There is a logger for console, which is send in production to apache and everything that is a warning or higher should be sent to Sentry.

@Phyks Since you recently worked Sentry support, could you do that for the logs to? I Think that in `dissemin/settings/common.py` should be everything ready according to this documentation: 
https://docs.sentry.io/platforms/python/logging/#handler-classes
https://forum.sentry.io/t/sentry-python-dictconfig/5065/3

Also, not all print statements where removed, those from the tests, migrations and settings are still remaining:

```bash
$ grep -rn -e "\(\$\| \)print(" ./ --include=*py
./deposit/tests/test_protocol.py:131:            print(form.errors)
./deposit/tests/test_protocol.py:149:            print(self.proto._logs)
./dissemin/settings/common.py:60:        print('Sentry module is not available although a Sentry DSN was set. '
./papers/migrations/0036_paper_temp.py:19:        print(lastpk)
./papers/migrations/0004_strip_latex.py:19:                print('"%s" -> "%s"' % (paper.title,new_title))
./papers/tests/test_ajax.py:52:            print("Invalid status code %d, response was:\n%s" %
./papers/tests/test_pages.py:64:            print(resp.content)
./papers/tests/test_pages.py:65:            print("HTML validation error: "+str(e))
./papers/tests/test_pages.py:211:                print(p)
./papers/tests/test_bibtex.py:124:        print(parse_bibtex(bibtex)['author'][0][0])
./papers/tests/test_orcid.py:45:            print(e)
./papers/tests/test_orcid.py:48:            print(r.json())
./upload/tests/test_views.py:102:            print("Invalid status code %d, response was:\n%s" %
./upload/tests/test_views.py:109:            print(("Invalid status code %d, response was:\n%s" %
```

while the rest is replaced with logging statements.

No new logging statements where added.

### DIFF
--- a/backend/crossref.py
+++ b/backend/crossref.py
@@ -23,6 +23,7 @@
 import datetime
 import json
 import bz2
+import logging
 
 import requests
 from requests.exceptions import RequestException
@@ -55,6 +56,8 @@ from publishers.models import Journal
 from publishers.models import Publisher
 from backend.pubtype_translations import CROSSREF_PUBTYPE_ALIASES
 from time import sleep
+
+logger = logging.getLogger('dissemin.' + __name__)
 
 ######## HOW THIS MODULE WORKS ###########
 #
@@ -193,9 +196,8 @@ def create_publication(paper, metadata):
     """
     try:
         return _create_publication(paper, metadata)
-    except DataError as e:
-        print("create_publication: ignored DataError:")
-        print(e)
+    except DataError:
+        logger.exception("Ignored DataError")
 
 
 def _create_publication(paper, metadata):
@@ -312,9 +314,8 @@ def fetch_dois_incrementally(doi_list):
     for doi in doi_list:
         try:
             metadata = fetch_metadata_by_DOI(doi)
-        except MetadataSourceException as e:
-            print("MetadataSourceException ignored:")
-            print(e)
+        except MetadataSourceException:
+            logger.exception("MetadataSourceException ignored")
             continue
         yield metadata
 
@@ -405,7 +406,7 @@ class CrossRefAPI(object):
             try:
                 p = self.save_doi_metadata(metadata)
             except ValueError as e:
-                print(e)
+                logger.info("Returning 'None'", exc_info=e)
         return p
 
     def save_doi_metadata(self, metadata, extra_orcids=None):
@@ -524,7 +525,7 @@ class CrossRefAPI(object):
                 if not found:
                     break
                 next_cursor = jpath('message/next-cursor', js)
-                print('Next cursor: '+next_cursor) # to ease recovery
+                logger.info("Next cursor: " + next_cursor) # to ease recovery
             except ValueError as e:
                 raise MetadataSourceException(
                     'Error while fetching CrossRef results:\nInvalid response.\n' +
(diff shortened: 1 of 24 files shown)


wetneb:
Thank you very much for this Stefan! If you need help with making Travis green again let me know.

beckstefan:
I'm on investigation. It has something to do with raven, which is somehow installed on my development system.

At least I found out that way, that my sentry approach and link above is deprecated. I'll try to get this done.

dissemin/settings/common.py
```diff
@@ -374,22 +374,31 @@
...
             'level': None,
-            'handlers': ['console', 'sentry'],
+            'handlers': ['console'],
             'propagate': False,
         },
     },
 }
+
+# If sentry is set, we send all important logs to sentry.
+SENTRY_DSN = True
```
Phyks: Isn't this a leftover from testing?

backend/crossref.py
```diff
@@ -558,9 +559,9 @@ def fetch_and_save_new_records(self, starting_cursor='*', batch_time=datetime.ti
                     p = Paper.from_bare(bare_paper)
                     p.update_index()
                 except ValueError as e:
-                    print((record.get('DOI') or 'unknown DOI') + ': '+str(e))
+                    logger.info(record.get('DOI', 'unkown DOI') + ': %s' % e)
```
Phyks: What is the decision with regards to having `exc_info` or exception text added in the string? (see `exc_info` above)

Phyks: Or `logger.exception` :)

wetneb: I think `info` should be right here in terms of severity level. This is something that routinely happens when fetching bogus metadata from crossref (papers with no authors, papers with the abstract in the title field, and so on).

dissemin/settings/common.py
beckstefan: Yep, thanks.

backend/crossref.py
beckstefan: The above construction just gives the exception message but not the traceback / exc_info, i.e. looks as before:
`2019-03-22 10:01:54 INFO dissemin.backend.crossref:562  10.1007/s35152-018-0026-3: Invalid author`
I think: If this is reported, there should be reason, i.e. what is wrong with the metadata of the DOI.

beckstefan:
Well, I do not have any experience with sentry yet, so I do not know if implementing the documentation works as expected.

backend/crossref.py
Phyks: Ok!

Phyks:
> Well, I do not have any experience with sentry yet, so I do not know if implementing the documentation works as expected.

Neither do I ^^ And the Sentry doc can be quite confusing from time to time. I just played a bit recently with it to have JS integration etc. @RaitoBezarius probably has more knowledge about this.

The current version looks fine to me, I think we can merge it and rework afterwards if we see any Sentry-related issue.

Thanks a lot for the work on changing logging mechanism!

Issue #629, opened 2019-02-28
Names links on profile does not work


Phyks:
On an author profile page, there is a list of alternative names (disambiguation). I think this is only shown when logged in as an admin.

Typically, for my profile, this looks like

![capture d ecran 2019-02-28 a 11 33 52](https://user-images.githubusercontent.com/3856586/53560526-da727c00-3b4c-11e9-8da8-9dbc95aec53d.png)

Clicking the link redirect to `https://dissem.in/search/?authors=<AUTHOR_ID>` and it does not work. Not sure what should be done here, but I guess this should either be a search on `AUTHOR_ID` or a fulltext search on disambiguated author name.

wetneb:
Yes we should remove these names - I will actually remove the corresponding model altogether, I think, as this is a leftover from the previous author name disambiguation heuristics.

Phyks:
Same for metadata links (in https://dissem.in/p/9364/hybridization-of-first-and-second-sound-in-a-weakly-interacting-bose-gas for instance) after decommissioning doai.

wetneb:
Fixed by #664.

PR #664, opened 2019-03-26
Remove the NameVariant model, unused

wetneb:
This is mostly to clean up the researcher profiles as seen by superusers. This model was introduced when we had an author disambiguation algorithm and is no longer needed, as papers are attributed via ORCID or manually.


Phyks:
Note: This will likely fix https://github.com/dissemin/dissemin/issues/629.

PR #672, opened 2019-03-27
Date on paper page

### LINKED ISSUE #633: Paper page should display publication date
The paper page currently does not display the publication date (or year):
https://dissem.in/p/99648067/carvacrol-rich-oregano-oil-and-thymol-rich-thyme-red-oil-inhibit-biofilm-formation-and-the-virulence-of-uropathogenic-escherichia-coli

We should add it somewhere!

beckstefan:
Adds the year in front of authors list as "Article published in 2019 by Max Mustermann"

This PR changes some po-files, should I commit these too or is there another workflow for dissemin?

Closes #633

### DIFF
--- a/papers/templates/papers/paper.html
+++ b/papers/templates/papers/paper.html
@@ -31,122 +31,124 @@
 {% endblock %}
 
 {% block jsScript %}
-$(function(){
-$('.helpPopover').popover({trigger: "hover"});
-});
-{% if paper.has_many_authors %}
-$(function(){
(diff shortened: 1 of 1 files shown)


wetneb:
@beckstefan great!

About the `.po` files, we should indeed decide on a policy for that. I think it is fine not to update the `.po` files in the PR, as it makes it easier to review (given that we have quite a lot of languages now). We can update them separately from time to time.

PR #674, opened 2019-03-27
Better depositing UX

### LINKED ISSUE #382: Upload interface: "deposit" button is enabled although mandatory information is missing, and an error message is lacking
When uploading to HAL, the "deposit" button appears as clickable even though all the required fields are not filled. If some fields are missing, the click produces no effect but no error message is displayed either. 

Deposit button should not appear as clickable (greyer / paler color ?) and an informative error message ("required fields in the metadata are missing") should be displayed. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43881724-upload-interface-deposit-button-is-enabled-although-mandatory-information-is-missing-and-an-error-message-is-lacking?utm_campaign=plugin&utm_content=tracker%2F6475570&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F6475570&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #657: Clickable regions on the deposit page
On the deposit page:

- [x] Clicking the description of a repository should select the radio button for that repository
- [x] Clicking the title bar of a panel should expand/collapse it (not just the text in the title bar)

Phyks:
Should fix #657 and fix #382.

Remaining point is that the JS validation of the form is not yet done. Actually, the form is validated server side with Django, it's a bit more work to handle this. We can open a separate issue to track this, it would probably require using http://emadmokhtar.com/client-side-validation-for-django-forms.html.

### DIFF
--- a/deposit/static/css/deposit.css
+++ b/deposit/static/css/deposit.css
@@ -74,9 +74,12 @@ label[class="radio"] {
     margin: auto;
 }
 
-.repositoryItem {
+label.repositoryItem {
+    display: block;
+    margin-bottom: auto !important;
     border-top: 1px solid #CCC;
     padding-right: 10px;
+    font-weight: normal;
 }
 
 .disabledRepositoryItem p {
--- a/deposit/templates/deposit/start.html
+++ b/deposit/templates/deposit/start.html
@@ -34,7 +34,7 @@ yl.jQuery = $;
 $(function() {
     $('#waitingArea').hide();
     $('#uploadFileId').val('');
-    
+
     /* Prefetched fields */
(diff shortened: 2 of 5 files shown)


Phyks:
Actually, I did a bit of client side basic jQuery code to handle the verification in https://github.com/dissemin/dissemin/pull/674/commits/ce722d0e0e408172f4f4e5dcade083c4cbfd154f. This seems to work well so far :)

Issue #661, opened 2019-03-26
Main page search form


Phyks:
Hi,

I was thinking about the search form on the index page. I recently discovered inadvertently that the search form supported more complex syntax than just "First Last". I don't see documentation on this (and discoverability would be better anyways) and I don't think a user could discover it on its own.

The currently supported syntax is (FIRST and LAST are placeholders):

* "FIRST LAST" (or "LAST FIRST", see #516 and #642).
* "last:LAST" to enforce last name.
* "ORCID_ID" to enforce an ORCID Id.
* Comma for enforcing separation of the names.

First question is why don't we support the "first:FIRST" counterpart of "last:LAST"? If my name is "Martin" for instance, I guess the impact of enforcing it to be last name, but this is likely to not be enough to further filter.

We might be able to handle it in a similar way as Github, with a link below the search input "See help for advanced search syntax" linking to a FAQ entry similar to https://help.github.com/en/articles/searching-issues-and-pull-requests.

What do you think?

beckstefan:
Sounds very good.

I don't know if makes sense for this search, but people are used to quotes from Google Search, i.e. to enforce to search for "Martin Daniels" as phrase they will type `"Martin Daniels"`.

As a bonus we could add on the start page with the search field an advanced search button to make user input easier than reading documentation.

Phyks:
Thinking about it, supporting both commas and quotes is going to be a nightmare (and probably also for users). So we should probably drop commas as separators if we want to support quotes. I don't have very strong opinions here, I'd tend to think commas are easier to handle and already here (so we have backward compatibility supporting them).

For the rest of the issues discussed here, I have an incoming PR.

Phyks:
Actually, supporting both first: and last: would probably require quotes (and much more search logic).

For instance, if I want to search for `first:"Paul André" last:Dupont`. I'd rather be in favor of keeping the statusquo so far and maybe have a separate issue tracking improved search logic if there is a need for it (is there?).

a3nm:
Naive opinion here: advanced search operators in the search field of the main page are not easily discoverable (I didn't know about them either) and I don't expect that many people will use them.

My opinion would be that it's fine not to bother with them too much (e.g., make the simplest possible choices in terms of parsing), and encourage users that want to make complicated queries to use the advanced search page. (In particular when the user submits a query with advanced operators that cannot be parsed properly.)

Issue #678, opened 2019-03-28
Missing form entries in the right hand side menu


Phyks:
![Capture d’écran 2019-03-28 à 11 18 18](https://user-images.githubusercontent.com/3856586/55149859-3d0a6800-514b-11e9-8926-e58492913ba9.png)

Issue #354, opened 2017-03-03
"By publication date" interface is unclear [$10]


a3nm:
On http://new.dissem.in/search the fields to search "By publication date" are not clear. They are free-form text inputs, and one doesn't know which kind of date format is expected in them.

PR #684, opened 2019-03-28
Ideas for a rework of main search form

Phyks:
Looking at https://dissem.in/, I was thinking that:

- The Dissemin logo is never used on the website except for the favicon. Therefore, it might not be well associated with Dissemin by users.
- The "Search" button lacks contrast with the rest of the search bar.
- We sometimes use "dissemin" (small letters) and sometimes "Dissemin" (capitalized).
- The search form might not allow easy discovery of all of its possibilities. See #661.

Here is a quick draft of a rework, let me know what you think about it. 

Screenshot_2019-03-28 Accueil - Dissemin

Some questions I have so far:

- [x] Is "Dissemin" or "dissemin" the correct/expected spelling?
- [x] There are many different logos out there, all different. The one from CAPSH website https://association.dissem.in/files/grue_blue_hres.png has a light blue, different from the dissemin favicon https://dissem.in/static/favicon/android-icon-192x192.png and the Twitter one seems to be even darker https://pbs.twimg.com/profile_images/636410533399425024/svih65R2_400x400.png. Which one is the current one? We should probably homogeneize this.


wetneb:
About dissemin vs Dissemin, no strong feelings for me (we used to try to enforce the lowercase initially, I think, but in the end I find it more natural with a capital letter as it is visually clearer that it is a proper noun). I agree with all other changes.

Phyks:
No longer a WIP. Here is the result.

Index page:

Screenshot_2019-03-29 Accueil - Dissemin

New advanced search page:
Screenshot_2019-03-29 Advanced search - Dissemin(1)


Search results with the form:
Screenshot_2019-03-29 Papiers - Dissemin


This fixes #661 as well as #678 and #354.

Phyks:
Note, the only question remaining (on my side) is the logo question (different logos with different background colors).

wetneb:
@Phyks aaawwww. Jaw droppingly amazing.

Concerning the logo, I am not sure! Is there a version you prefer? We can decide it is the official one and replace the others elsewhere.

Phyks:
I don't like much the current one in the favicon, I feel like the blue background does not connect with anything Dissemin related (dark blue, but not the same one as the dominant blue on dissem.in).

The Twitter one is the same blue as the one we already use everywhere (CAPSH/Dissemin etc). Maybe this is the best choice?

Fwiw, here is how it looks like in the navbar:

Capture d’écran 2019-03-29 à 12 26 18

The light blue one from CAPSH website looks nice also in the navbar, but this should not be the reason to choose it as the logo conveys much more identity than just its effect in the navbar.

Capture d’écran 2019-03-29 à 12 22 01

a3nm:
No strong opinion about the logo; I agree it's a good idea to include it on the website. The rest of the rework looks great to me. :)

Phyks:
I pushed an extra commit to update the favicons.

For https://association.dissem.in/files/grue_blue_hres.png, it could be replaced by 
grue_blue_hres

wetneb:
@Phyks great!!! Feel free to merge when appropriate and I'll be very excited to deploy it :)

Phyks:
Thanks! I'll probably not be much available until Monday, maybe not the safest thing to do to deploy right now :p

beckstefan:
Couldn't we do the last:Doe first:John put into the advanced search in doing separate fields for last and first name? This would not require any quotes, I think.

And most important: Big thumbs up from me!

Phyks:
> And most important: Big thumbs up from me!

Thanks!


> Couldn't we do the last:Doe first:John put into the advanced search in doing separate fields for last and first name? This would not require any quotes, I think.

It's slightly more difficult than this, since this field can be used to look for multiple authors. So it would mean having separate last/first names. I'm not sure this would be much easier for users then and I'd personnally rather try to keep a single field as much as possible.

beckstefan:
Ah, I see. Thanks for the explanation.

Would it work to dynamically add new author name fields? I mean, per standard one author with fields for last and first name and behind that a Plus-Button/Add-Author-Button to add another author, which generates a new set of a tuple for last and first name, via JS? So then it would be clear which first name corresponds to which last name.

Phyks:
Updated on http://association.dissem.in/. Let's merge this one.

I created https://github.com/dissemin/dissemin/issues/696 to track the discussions on the UX of the search field, which would deserve a bit more work in my opinion.

Issue #659, opened 2019-03-25
Papers with ORCID ids are not linked to Researcher objects


wetneb:
When refetching a Researcher profile, it can be the case that there are papers not linked to that Researcher, but which contain their ORCID id in the authors list. In that case, we want to find these and link them to the Researcher object. For some reason that does not seem to be happening at the moment.

wetneb:
Example:
- https://dissem.in/r/142736/nick-fox
- https://dissem.in/p/103879789/seizures-as-an-early-symptom-of-autosomal-dominant-alzheimers-disease
Refetching the first does not associate the second.

wetneb:
Actually, we don't index ORCID ids in papers at all! Simply scandalous. But easy to fix.

wetneb:
Fixed by #694.

PR #694, opened 2019-03-31
Index ORCID ids in Paper objects

wetneb:
This changes the search index so that ORCID ids are directly indexed there.

We currently only index researcher ids (we don't have researchers for every ORCID profile, hence the discrepancy).

Tests are still needed and `link_existing_papers` needs to be hooked to the rest of the code for #659.

### DIFF
--- a/backend/orcid.py
+++ b/backend/orcid.py
@@ -39,9 +39,11 @@ from papers.baremodels import BarePaper
 from papers.errors import MetadataSourceException
 from papers.models import OaiSource
 from papers.models import Researcher
+from papers.models import Paper
 from papers.orcid import OrcidProfile
 from papers.orcid import affiliate_author_with_orcid
 from papers.utils import validate_orcid
+from search import SearchQuerySet
 
 logger = logging.getLogger('dissemin.' + __name__)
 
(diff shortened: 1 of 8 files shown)


wetneb:
This is now ready. After deploying this, we will need to recompute the ES index entirely.

backend/orcid.py
```diff
@@ -146,6 +149,35 @@ def warn_user_of_ignored_papers(self, ignored_papers):
...
+                new_authors = paper.authors
+                for author in new_authors:
+                    if author.orcid == researcher.orcid:
+                        author.researcher_id = researcher.id
+                paper.authors_list = [author.serialize() for author in new_authors]
+                papers_to_update.append(paper)
+                paper.update_index()
+        
+        if papers_to_update:
+            if django.VERSION[0] > 2 or (django.VERSION[0] == 2 and django.VERSION[1] >= 2):
```
Phyks: Is there a main reason to support multiple Django versions?

Phyks:
Eager to see this landing in production, and hoping that both #730 and related will be fixed :)

backend/orcid.py
wetneb: Yeah let's just require 2.2, this `bulk_update` thing is going to be useful at many other places I think

Issue #685, opened 2019-03-29
Bug with changing search ordering


Phyks:
Quite a blocking bug actually, when searching by "title" ordering rather than "date", the search results are empty :/

Compare https://dissem.in/search/?authors=amarilli&sort_by=text with https://dissem.in/search/?authors=amarilli for instance.

Phyks:
Actually, sorting by title is not very useful, this sorting possibility should probably be discarded.

Issue #695, opened 2019-04-01
ElasticSearch caching causes memory error


wetneb:
From Sentry: https://sentry.io/organizations/dissemin/issues/963517511/?project=189330&query=is%3Aunresolved&statsPeriod=14d&utc=false

Some large search queries seem to generate large mallocs because of https://github.com/django-haystack/django-haystack/issues/1219. Curious that we haven't stumbled upon this before! Given the size of our index this is absolutely critical.

Phyks:
What do you think about https://github.com/django-haystack/django-haystack/issues/1219#issuecomment-483259635?

The solution could be ok, I think, so we could actually rebase this PR and give it a try?

wetneb:
I suspect this issue is the cause of our current performance issues, so I will try deploying this fix in production (YOLO) and see what happens.

PR #697, opened 2019-04-01
First draft of an export citations feature

### LINKED ISSUE #688: Export list of papers
Now that we have Bibtex export of citations for individual papers, we could add export for list of papers.

On researcher page, it would be fairly easy to add export of all the papers from the researcher as BibTeX.

On search page, I've mixed feelings about this. I'd like to add the ability to export search results to BibTeX. The issue is that with a wide search, exporting 1M papers to BibTeX is simply not realistic. Not sure about the best approach here:
* Only offer BibTeX export for searches with few enough results (but this is a bit arbitrary and might be difficult to understand for the users).
* Only offer BibTeX export of the current page (maybe the best solution).

What do you think about the search page? Would it be worth it?

Phyks:
* Export citations for search pages and researcher profiles pages.
* Move the sorting button out of the search sidebar.

The API part is intentionally very limited, this will be better handled later on when refactoring the API altogether I think.

Fix #688 and #685.

Capture d’écran 2019-04-01 à 16 40 05
Capture d’écran 2019-04-01 à 16 40 11
Capture d’écran 2019-04-01 à 16 40 17

Left to do:

- [x] Restore the sorting button logic.
- [x] Handle duplicate IDs in BibTeX.
- [x] Write tests

Currently not sure whether the export citations button on researcher page should export the whole profile citations or only the currently displayed page (current behavior).


Phyks:
Should be ready. Current version looks like

Screenshot_2019-04-02 Papiers - Dissemin

Screenshot_2019-04-02 Antoine Amarilli - Dissemin

Export button only exports the citations for the current page so far. What should be the expected behavior here?

Note that currently when changing the search fields, this resets the sorting order. I felt this was a bit more natural, but I'm not 100% convinced.

a3nm:
The button could be "Export this page in BibTeX"? or "Export citations for this page" if we have different output formats.

I didn't understand the part about resetting the sorting order?

wetneb:
A few thoughts about the layout:
At the moment the location of the sorting widget and the export button feel a bit too prominent on the page. Bibtex export is a great feature to have, but it is not necessarily the first thing we want to show to users.

Now that we have an advanced search form, I'd be in favour of trimming down the right-hand side: only showing the title and author search field by default, and the others only if they are in use (they have non-empty values). That would reduce the area taken by this search form, which feels a bit heavy to me: currently, the search button is really far down the page. We would add a link to the advanced search page next to the search button, so that people could still access the search fields not displayed by default. Also, this would give a purpose to the advanced search page, as currently it does not display more fields than the search interface itself.

If we can do this, then we could move the sort widget and the export button to the right hand side. Having the export there would be more consistent with the paper page, where export is already located there.

wetneb:
(Of course all of the above are quite some work and could be done outside this PR)

Phyks:
> I didn't understand the part about resetting the sorting order?

@a3nm Currently, say you have an empty search ordered by "oldest first", when narrowing down the search by authors, the search order will reset to "newest first".

> At the moment the location of the sorting widget and the export button feel a bit too prominent to me. Bibtex export is a great feature to have, but it is not necessarily the first thing we want to show to users.

@wetneb I definitely agree with it, and this was bothering me as well. I was thinking into turning the "Export" button into an "Action" button (or dropdown actually with a list of actions) to act on papers. But there are not much other actions coming to me.


> Now that we have an advanced search form, I'd be in favour of trimming down the right-hand side: only showing the title and author search field by default, and the others only if they are in use (they have non-empty values). That would reduce the area taken by this search form, which feels a bit heavy to me: currently, the search button is really far down the page. We would add a link to the advanced search page next to the search button, so that people could still access the search fields not displayed by default. Also, this would give a purpose to the advanced search page, as currently it does not display more fields than the search interface itself.

I don't like much having the search form changing number of fields based on the current search :/ What about having collapsible form groups?

Here is a quick draft of what I have in mind. This means:
- Changing the sidebar title to "Results"
- Adding new titles in the sidebar for "Tools" and "Refine search" sections
- Having all the tools (sorting order / citations export etc) in a dedicated section.
- Having the current search form in the same form in the "Refine search" section except only the fields titles would be shown (unless they have a non-empty value). A field with a non-empty value (or when clicking the caret) would be shown.

To get a hint of the resulting view, an iPad (1024x768) in landscape mode would basically have the search button in the initial view when only one field is filled in, I think. One a desktop computer the search button would likely be in the initial view for most common searches.

Capture d’écran 2019-04-02 à 17 29 20

wetneb:
That would be simply fantastic!

Phyks:
Great! I'll refactor it this way then :)

Phyks:
Quick update on this one, here is what I have so far:

Screenshot_2019-04-03 Antoine Amarilli - Dissemin
Screenshot_2019-04-03 Papiers - Dissemin

I still have a bit of work to ensure everything works fine with the new sort by buttons and OA statistics filtering.

Phyks:
This was a bit more complicated than expected to rework the forms :) It should almost be ready by now (at least for testing it).

I'll try to fix the tests and squash the commits for a cleaner history and easier review tomorrow.

papers/tests/test_api.py
```diff
@@ -126,3 +126,81 @@ def test_bibtex_formatting(self):
...
+  year = {2015}
+}"""
+        from django.core.management import call_command
+        call_command('clear_index', interactive=False)
+        r1 = Researcher.create_by_name('Antoine', 'Amarilli')
+        p1 = Paper.create_by_doi('10.1109/lics.2015.37')
+        p1.set_researcher(0, r1.id)
+        p2 = Paper.create_by_doi('10.1145/2744680.2744690')
+        p2.set_researcher(0, r1.id)
+        call_command('rebuild_index', interactive=False)
```
wetneb: At the moment, if you do not explicitly index papers in ElasticSearch with `p1.update_index()`, it is normal that they do not end up there. So I would expect that you should be able to replace `call_command` by `p1.update_index()` and `p2.update_index()`.

Phyks:
@wetneb I discovered there is a bad interaction with respect to the index in the tests :/

Minimal command to reproduce the failing test cases here is:

```
python -m pytest backend/tests/test_orcid.py papers/tests/test_pages.py
```

Actually, something bad happends in `backend/tests/test_orcid.py` and destroys the index. Here is a quick reproducing guide:

* With a working database and index (typically https://github.com/dissemin/dissemin/blob/master/provisioning/provision.sh#L119-L121), run `python -m pytest backend/tests/test_orcid.py`.
* Then, start the server with `python manage.py runserver 0.0.0.0:8080` and head to http://localhost:8080/search/.
* The page looks weird:

Capture d’écran 2019-04-05 à 14 06 32

This is because the index has been somehow broken by the `backend/tests/test_orcid.py` test. If you inspect the `object_list` used in templates (https://github.com/dissemin/dissemin/blob/master/papers/templates/papers/paperList.html#L9), this is a huge list of `None` values, probably linked to #695.

You can recover from this situation by running `python manage.py rebuild_index`.

This also means that in `papers/test/test_pages.py` we are not testing against search page listing papers but against the very weird search page from the above screenshot. :/

Phyks:
Seems that https://github.com/dissemin/dissemin/pull/697/commits/bdf34d6fb851e72c187b9928092cc66d63d8ff79 does the trick :)

PR #698, opened 2019-04-01
Fix lost onChange event on new date pickers

Phyks:


Phyks:
Should be mergeable. First commit addresses the fact that the new Bootstrap datepickers no longer emit a change event when edited.

Second commit addresses a weird bug I had with `window.history` functions ignoring empty query strings. The bug can be reproduced quite easily by going to https://dissem.in/r/4162/lucas-verney?q=escape and clearing the filter on the title.

PR #699, opened 2019-04-02
Update VM config to Stretch

Phyks:
<s>WIP status since I have to recheck full build.</s>

### DIFF
--- a/dissemin/settings/common.py
+++ b/dissemin/settings/common.py
@@ -192,6 +192,7 @@ TEMPLATES = [
                 ),
                 'context_processors': (
                     "django.contrib.auth.context_processors.auth",
+                    'django.contrib.messages.context_processors.messages',
                     "django.template.context_processors.debug",
                     "django.template.context_processors.i18n",
                     "django.template.context_processors.media",
--- a/dissemin/settings/dev.py
+++ b/dissemin/settings/dev.py
@@ -62,28 +62,7 @@ EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 MEDIA_ROOT = os.path.join(BASE_DIR, 'dissemin_media')
 
 # Disable caching in dev
-TEMPLATES = [
-    {
-        'BACKEND': 'django.template.backends.django.DjangoTemplates',
-        'DIRS': [
-            os.path.join(BASE_DIR, 'templates')
-        ],
-        'APP_DIRS': True,
-        'OPTIONS': {
-            'context_processors': (
-                "django.contrib.auth.context_processors.auth",
-                "django.template.context_processors.debug",
-                "django.template.context_processors.i18n",
-                "django.template.context_processors.media",
-                "django.template.context_processors.static",
-                "django.template.context_processors.tz",
-                "django.template.context_processors.request",
-                "django_settings_export.settings_export",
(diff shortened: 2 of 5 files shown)


dissemin/settings/common.py
```diff
@@ -192,6 +192,7 @@
                 ),
                 'context_processors': (
                     "django.contrib.auth.context_processors.auth",
+                    'django.contrib.messages.context_processors.messages',
```
Phyks: I got an error about Django admin missing this.

dissemin/settings/dev.py
```diff
@@ -62,28 +62,7 @@
...
-                "django.template.context_processors.request",
-                "django_settings_export.settings_export",
-                "dissemin.tcp.orcid_base_domain",
-            ),
-            'debug': True
-        }
-    }
-]
+del TEMPLATES[0]['OPTIONS']['loaders']
+TEMPLATES[0]['APP_DIRS'] = True
```
Phyks: `dev` config was overwritting the `TEMPLATES`, so I instead just overloaded it.

provisioning/provision.sh
```diff
@@ -59,16 +59,16 @@ systemctl restart elasticsearch
 # We install some dev tools (tmux and vim)
 apt-get install -y tmux vim-nox
 # We create a virtualenv for Dissemin
-python3 -m venv /dissemin/.vm_venv
+python3 -m venv /home/vagrant/.vm_venv
```
Phyks: I got weird issues with `pip` when running `--upgrade` in the VM, `EnvironmentErrors` and issues with deleting cache folders from pip (starting with a `~`). I guess this is because the virtualenv and the special characters was living in a shared folder, it seems to be fine with the virtualenv on the VM disk only.

provisioning/provision.sh
```diff
@@ -79,55 +79,9 @@ then
...
-# RoMEO API KEY
-# Used to fetch publisher policies. Get one at
-# http://www.sherpa.ac.uk/romeo/apiregistry.php
-ROMEO_API_KEY = None
-
-# Proaixy API key
-# Used to fetch paper metadata. Get one by asking developers@dissem.in
-PROAIXY_API_KEY = None
-EOF
+cp /dissemin/dissemin/settings/secret_template.py /dissemin/dissemin/settings/secret.py
```
Phyks: This should be initialized from template to avoid forgetting to update it :)

requirements.txt
```diff
@@ -2,7 +2,7 @@ celery[redis]
 pyoai
 PyPDF2
 bibtexparser>=1.1.0
-redis<3
+redis
```
Phyks: This one is tricky, I got an error that version 3.2.0 was required. This is coming from https://github.com/celery/kombu/pull/1011 (which is pulled as a dependency to another lib, maybe `celery[redis]`?).

@wetneb You forced a downgrade to `redis<3` a while ago (https://github.com/dissemin/dissemin/commit/cafc862da005ea25c816018d5d116f1d538d5f89), do you remember why? My change might be problematic.

wetneb: That's apparently because celery 4.3.0 was released. I think the previous version was not compatible with redis 3. So removing this should be fine now!

Phyks:
I rebuilt a VM from scratch, this should be fine :)

PR #700, opened 2019-04-02
Better print style

Phyks:


Phyks:
Current render in Chrome/Chromium looks nice:

* https://pub.phyks.me/tmp/home.pdf
* https://pub.phyks.me/tmp/search.pdf
* https://pub.phyks.me/tmp/researcher.pdf
* https://pub.phyks.me/tmp/paper.pdf

Printing paper lists in Firefox is still broken, I'm a bit puzzled on this one :/

Phyks:
So the issue in Firefox is actually linked to the use of `flex` items, see https://bugzilla.mozilla.org/show_bug.cgi?id=1089549 (5 years old issue, no activity :/).

Good news is that we can discard them easily I think, and it should work. I'll have a look.

Phyks:
Should be fine with Firefox now. At least, the list is correctly printed. The last item of the paper list however still seems to have issues. I think this one should be ready for a review and merge now :)

Phyks:
Should be rebased and squashed.

PR #701, opened 2019-04-04
Make OA Button a button

### LINKED ISSUE #653: Make the OA-Button a Button
Currently it is just some text.

Maybe we could write

> Not the author? Request a open access publication with _OAB_

beckstefan:
Closes #653 

Since there is some more text, I moved the button from the right box to the middle.

Currently I am not convenient with the button itself. I think a blue colour would be nice, but I don't really know how to make it blue (or whatever colour).

oa_button


papers/templates/papers/paper.html
```diff
@@ -146,6 +146,20 @@ <h1 id="paperTitle" data-params="{csrfmiddlewaretoken:'{{csrf_token}}'}" data-pk
...
+        {% if doi and not paper.pdf_url %}
+            <h2>{% trans "Not the author?" %}</h2>
+            <p>
+                {% blocktrans count author_nb=paper.author_count %}
+                    Encourage the author to liberate this publication with the Open Access Button.
+                    {% plural %}
+                    Encourage the authors to liberate this publication with the Open Access Button.
+                {% endblocktrans %}
+            </p>
+            <p><a href="https://openaccessbutton.org/request?doi={{ doi }}" target="_blank" class="btn btn-default btn-xs"><img src="{% static "img/icon_OAB.png" %}" style="height:20pt"> {% trans "Send request" %} <img src="{% static "extlink.png" %}" alt="[^]"></a></p>
```
Phyks: You can use `btn-primary` here instead of `btn-default` to make it blue I think. See https://getbootstrap.com/docs/3.3/components/#btn-dropdowns-single for instance.

Do you know if OAB has a specific policy for styling their logo?

Phyks:
Few extra things:

- maybe the external link indicator is not so useful on the button. I would expect the wording to imply enough that this is gonna trigger some external call.
- maybe the "Open Access Button" text could be a link to https://openaccessbutton.org/ to emphasize the previous point even more?
- I don't like so much the alignment here, maybe the button does not have to be put on a new line? Or maybe centering the text, like the bottom text in author page: https://dissem.in/r/4162/lucas-verney?

Thanks a lot for tackling this one! Looks really cool to get this out of the sidebar :)

wetneb:
Nice! I would suggest to add `rel="nofollow"` on the OAB link which includes a DOI, as it seems to dynamically fetch the DOI metadata on request (so we probably don't want bots to crawl these URLs).

papers/templates/papers/paper.html
beckstefan: Thanks!

Their site content is CC-BY. The only thing happening here is a resize (in the browser), that's fine.

wetneb:
I have dropped a line to @JosephMcArthur about this, to get his advice on the matter.

JosephMcArthur:
Hey folks! This is awesome thank you! I think adding our logo isn't required, to be honest.

beckstefan:
Ok, I removed the logo.

papers/templates/papers/paper.html
```diff
@@ -146,6 +146,20 @@ <h1 id="paperTitle" data-params="{csrfmiddlewaretoken:'{{csrf_token}}'}" data-pk
...
+        {% if doi and not paper.pdf_url %}
+            <h2>{% trans "Not the author?" %}</h2>
+            <p>
+                {% blocktrans count author_nb=paper.author_count %}
+                    Encourage the author to liberate this publication with the <a href="https://openaccessbutton.org/" target="_blank">Open Access Button</a>.
+                    {% plural %}
+                    Encourage the authors to liberate this publication with the <a href="https://openaccessbutton.org/" target="_blank">Open Access Button</a>.
+                {% endblocktrans %}
+            </p>
+            <p><a href="https://openaccessbutton.org/request?doi={{ doi }}" target="_blank" rel="nofollow" class="btn btn-primary btn-xs"><strong>{% trans "Send request" %}</strong></a></p>
```
Phyks: I'm not 100% convinced by the `btn-xs`, I feel like it lacks a bit of padding. What do you think?

Phyks:
In case this can be helpful, here is a quick draft of what I mentionned above.

Capture d’écran 2019-04-04 à 14 15 19

Here are my resulting edits:

```
<p style="text-align: center;" class="notFound">
  <strong style="color: #666;">Not the author?</strong> Encourage the authors to liberate this publication with the <a href="https://openaccessbutton.org/" target="_blank" class="btn btn-primary">Open Access Button</a>&nbsp;!
</p>
```
```

**Comment by wetneb** [comment]
It's minor but "Not the author? Encourage the authors […]" feels a tiny bit redundant because of the repetition of "author", especially now that the two are on the same line. Also I am not entirely sure that "Not the author?" is clear enough, I suspect many users would not realize it means "Are *you* not the author of this publication?".

I am not sure what wording I would suggest though. How about simply removing the "Not the author", maybe? The rest of the sentence is clear enough for people to realize that this feature should not be used by the authors themselves, I think.

**Comment by a3nm** [comment]
I agree that removing "Not the author" would be better than the current proposal.

Ideally I think it would be best to structure the page so that both options are clear, e.g.,

- *Are you the/an author of this publication?* Deposit it in an open repository... (links to deposit the various versions, policy details, etc.) Or maybe: *Do you have permission to deposit this publication* or just *Do you wish to deposit...*
- *Otherwise*, encourage the authors to deposit this publication using the OAB

**Comment by beckstefan** [comment]
I can do a proposition on Monday with a more clear structure.

**Comment by beckstefan** [comment]
A little late, but with two versions. I prefer the first one as it interferes better with a possible abstract.

Screenshot_2019-04-11 Neurofuzzy Decision-Making Approach for the Next Day Portfolio Thai Stock Index Management Trading St

Screenshot_2019-04-11 Neurofuzzy Decision-Making Approach for the Next Day Portfolio Thai Stock Index Management Trading St (1)

**Comment by Phyks** [comment]
Nice! I like the first one too :)

**Comment by wetneb** [comment]
Sorry to be a bit downbeat on this but I'm a bit reluctant with both actually.

The problem with the first one for me is that it presents the "email" option on the same footing as the article versions, which I find confusing - an email is not a paper version, right? So the green dot on the left does not mean much either - people could think this green dot comes from SHERPA/RoMEO due to the acknowledgement underneath.

The second removes the upload buttons next to the paper versions, which means that the deposit form will be longer (and it will require an extra click) as the paper version will have to be selected there.

My gut feeling is that this button does not need to be in the main text, but if you all think it is better like that of course I am happy to change this :)

**Comment by beckstefan** [comment]
Yes indeed. At first the _Contact author_ line was under SHERPA/RoMEO which is not very well placed (for the reason you name) and then I got lost in design, I think.

In the meanwhile I would state that the paper page is focussed on the publication and the upload, so after all the choice having the link in _Tools_ section is not that bad.

**Comment by Phyks** [comment]
> The problem with the first one for me is that it presents the "email" option on the same footing as the article versions, which I find confusing - an email is not a paper version, right? So the green dot on the left does not mean much either - people could think this green dot comes from SHERPA/RoMEO due to the acknowledgement underneath.

I was at first as well, but waited a couple of hours and changed my mind in the end. Still, this is true that the green dot could be confusing.

Why not putting it next to the "Fulltext: Unavailable" line?

Something like

Sans titre

With a button label like "ask authors" or "contact authors" maybe?

**Comment by Phyks** [comment]
Decision from dev meeting:

 which was the last proposal is nice, we should probably just invert "Data provided by Sherpa/Romeo" line and the "Contact authors" line to make two separate blocks clear.

File: papers/static/style/style.css [hunk 3]
```diff
@@ -527,6 +527,11 @@ h1#paperTitle {
     background-image: url('../question-circle.png');
 }
 
+.policyPic-ask
+{
+    background-image: url('../blue-circle.png');
```
beckstefan: In case of approval we need a bigger blue circle. I tried to reuse the other svgs to derive a blue circle, but wasn't succesful.

beckstefan:
Now it looks like this:

Screenshot_2019-05-03 Neurofuzzy Decision-Making Approach for the Next Day Portfolio Thai Stock Index Management Trading St

The blue circle is to small, see above.

papers/static/style/style.css
Phyks: "blue circle" is not a very nice name for this file, it is intended to be a dot for lists (`<ul>` type lists, small dot and not centered).

I think you can use https://pub.phyks.me/tmp/blue-circle.svg and https://pub.phyks.me/tmp/blue-circle.png for the circle here.

beckstefan: Thanks! I renamed the blue circle to a dot and added your new circles.

Phyks: Nice! Note that if you rename the previous file to a dot, you should also update the code using it currently, as it is expected the dot file and not the new circle one :/

beckstefan:
And now with a _real_ circle! :-)

Screenshot_2019-05-03 Neurofuzzy Decision-Making Approach for the Next Day Portfolio Thai Stock Index Management Trading St (1)

papers/static/style/style.css
Phyks: I commented too fast, before seeing you added a commit. Everything sould be fine, sorry for the noise :)

a3nm:
This looks great, thanks!

PR #709, opened 2019-04-12
Fix authors search tooltip alignment

### LINKED ISSUE #702: Search tooltip misaligned
The tooltip content is misaligned. The bullets are left aligned while their text is centered. This looks a bit weird and confusing to me.

search_tooltip

(Maybe brighter bullets would be nice.)

Phyks:
@beckstefan What do you think about this?

Capture d’écran 2019-04-12 à 14 04 08

Should fix #702.


beckstefan:
Nice!

Could you write _Use commas as name separator_? Sound's much nicer.

Phyks:
Done. Thanks!

Issue #563, opened 2019-01-25
Update frontend dependencies


Phyks:
Hi,

Dissemin uses a few JS libraries: JQuery (and a few modules), D3.JS etc. Some of them are in outdated version and we should probably update them.

It could probably be worth having a `package.json` file defining the requires dependences (for easier version management of the libs) and compiling all the vendor files into a single minified script.

Still, this is low priority for now :)

Phyks:
As a side note, it seems we are loading jQuery multiple times from different sources etc, might be worth investigating (on depositing page for instance).

PR #712, opened 2019-04-12
Move language toggle to navbar

### LINKED ISSUE #649: Fix Django-autocomplete-light used in the HAL interface
We currently have a JS error on the deposit page if the HAL repository is enabled.

We need to apply this fix: https://github.com/yourlabs/django-autocomplete-light/issues/1079#issuecomment-459725835

We might need to fork DAL and point our `requirements.txt` to that fork. Sad because we currently only point to official releases!

Once this is done, we can re-activate HAL (in the admin interface) and mark the sentry issues as resolved.

beckstefan:
Moves the language select up to the navbar, so that no one needs to scroll for his preferred language.

Screenshot_2019-04-12 Start - Dissemin

Screenshot_2019-04-12 Start - Dissemin(1)

Fixes #649.


Phyks:
I definitely agree with the removal of the "Home" link which is a duplicate of the left logo in the navbar. Having language selector in the navbar makes sense.

My only concern is that I feel like the selector is a bit too prominent at the moment. We can probably make it semi-transparent and shorten it. An example would be

Capture d’écran 2019-04-12 à 18 23 00

or



I like quite a lot the second one, which should not be much difficult with our Bootstrap CSS base.

Using flags is a known issue for language selector (typical example is French is not France only, language usually don't match with a whole and single country). If we want to use less space in the navbar, we could maybe use short locale ("FR", "EN", etc) forms (not sure there is much point in shortening though). The icon used by wikipedia is quite nice as well to mean localization https://fr.wikipedia.org/w/extensions/UniversalLanguageSelector/resources/images/compact-links-trigger.png?b0c8e.


Let me know what you think about these, I can give a hand on this if you want.


P.S.: Thinking out loud, are we detecting locales (from browser locale) automatically in Dissemin (I don't have the code base at hand to check right now)?  Otherwise, we can use something like https://framagit.org/phyks/cygnal/blob/master/src/i18n/index.js#L26-49 in JS.

wetneb:
This is all great! I agree we could do with a bit more styling as suggested by Phyks given that the selector is more visible now.

beckstefan:
> P.S.: Thinking out loud, are we detecting locales (from browser locale) automatically in Dissemin (I don't have the code base at hand to check right now)? Otherwise, we can use something like https://framagit.org/phyks/cygnal/blob/master/src/i18n/index.js#L26-49 in JS.

Django does that for us.

> Let me know what you think about these, I can give a hand on this if you want.

I like the second one. Although, we should take care of mobile users, so the dropdown should have it's own scrolling so that is does not eat up to much space, i think. 

Please feel free to push any changes to this PR!

Phyks:
Hi,

I rebased this branch and had a look at the style. Here is what I came up with so far.

* Desktop version: https://pub.phyks.me/tmp/desktop.mp4
* Desktop version without JS: https://pub.phyks.me/tmp/desktop_no_js.mp4
* Mobile version: https://pub.phyks.me/tmp/mobile.mp4

Let me know what you think about this (this branch is up to date if you want to test it live). If this is ok, I'll clean my code and merge this.

wetneb:
Amazing!

beckstefan:
Absolutely great!

Just two simple remarks: 
* Currently we host most of our third party scripts, why is select2 an exception?
* Desktop JS would be nicer with scroll bar on the left (might then be not good with Mobile)

Phyks:
> Currently we host most of our third party scripts, why is select2 an exception?

This is what I meant with "cleaning the code" :p (`select2` is used here and as part of `dal` for HAL form, I'll try to avoid duplication and bad interactions).

> Desktop JS would be nicer with scroll bar on the left (might then be not good with Mobile)

Like this?

Capture d’écran 2019-05-02 à 14 54 33

I tend to find it a bit unusual. What do you think?

beckstefan:
Yeah, the scroll bar looks very thick or so, indeed. But at least it's not an top of any language (which is unusual, too ;-))

Phyks:
Oops, sorry I got what was wrong with the first picture. I should have carefully reviewed the video!

Here is the real look on desktop device: https://pub.phyks.me/tmp/desktop_real.mp4.

I was using "responsive view" from Firefox devtools which puts a weird (mobile-like) scrollbar, this should not be normal behavior.

beckstefan:
Thanks for clarification!

I expected some JS/CSS-magic behind that and considered that as normal or as price. I could have indeed written what exactly wasn't in my favour. Next time!

Phyks:
Updated with locally serving the select2 files. I was afraid of a bad interaction with the select2 files from django-autocomplete-light (used for HAL repository). DAL is a nightmare to edit and fix and the only issue I could find was a bad sizing of the translation select on this specific page (deposit page):

Capture d’écran 2019-05-03 à 15 47 01

I would say this is good enough for now and mergeable. We'll see later as part of #563 which strategy we should consider with respect to DAL.

wetneb:
@Phyks is there no risk of incompatibility between DAL's select2 and ours?

a3nm:
Not very important but on mobile https://pub.phyks.me/tmp/mobile.mp4 I find it a bit confusing that the language is the first entry in the menu. I would have expected it to be, e.g., the last one after some kind of separator. What do you think?

Phyks:
> is there no risk of incompatibility between DAL's select2 and ours?

I was fearing this as well… DAL is only used for HAL form (I think) and therefore DAL's select2 is only loaded on the deposit page. The only bad interaction I found there is the incorrect (too large) width of the translation select2: https://github.com/dissemin/dissemin/pull/712#issuecomment-489107834.

I wish there was a way to prevent DAL from loading its select2 and use our own, but DAL is doing a lot of black magic and I'm not sure this is possible at all (and saying their doc is not very complete is a euphemism… :/). 

Looking at it a bit more in depth (and given that DAL is already a huge pain and we had to monkey patch it because of https://github.com/dissemin/dissemin/issues/649), I am thinking it might be better to ditch it and replace it with a lighter alternative.

Looking quickly online, I found https://docs.djangoproject.com/en/2.0/releases/2.0/#django-contrib-admin (not sure exactly what this implies, it seems Django somehow supports select2 fields). A promising alternative is https://github.com/applegrew/django-select2 which seems enough for our use case and is easily configurable through Django's settings: https://github.com/applegrew/django-select2/blob/f6eaa26c16eab84c8123a2760dcf9272f4dde0fd/django_select2/conf.py#L59-L93.

What do you think?


> Not very important but on mobile https://pub.phyks.me/tmp/mobile.mp4 I find it a bit confusing that the language is the first entry in the menu. I would have expected it to be, e.g., the last one after some kind of separator.

I agree. However, this would require using CSS flexboxes and `order` property (https://developer.mozilla.org/fr/docs/Web/CSS/CSS_Flexible_Box_Layout/Ordonner_%C3%A9l%C3%A9ments_flexibles#La_propri%C3%A9t%C3%A9_order) (the only way I know to have an ordering of elements which does not correspond to the order of apparition in the source code).

This is totally doable, the issue being IE6-9 (as always): https://caniuse.com/#feat=flexbox. We are currently only using flexbox in a very parcimonious way. Bootstrap did the jump to flexbox in the version 4 (we are using version 3), this could totally be doable for Dissemin in my opinion.

wetneb:
@Phyks ok, makes sense! Yes DAL is painful to work with, happy with changing to something else. This does not have to be part of this PR as long as the select widget in the HAL deposit interface remains functional on the short term.

Phyks:
So, DAL should be gone and replaced by Django-Select2. No more bad interaction possible between our Select2 code and the one from the lib.

I also implemented the `flexbox` solution for changing the ordering of items in the navbar and language selector in the mobile view.

Should be ready for a review and merge :)

PR #714, opened 2019-04-15
Tell sentry current commit

### LINKED ISSUE #608: Set up deployment tracking in Sentry
Sentry makes it possible to track when code is deployed and match that with the issues they detect.
We should look into adding this to dissemin. The main difficulty seems to be to expose the version / commit id to the Python code, but I'm sure there are standard ways to do that in Django projects.

https://docs.sentry.io/workflow/releases/?platform=python#configure-sdk

beckstefan:
Closes #608 

I am not sure if the commit itself is enough for the tracking. The sentry documentation is not very clear about that.

### DIFF
--- a/dissemin/settings/common.py
+++ b/dissemin/settings/common.py
@@ -31,6 +31,7 @@ https://docs.djangoproject.com/en/1.8/ref/settings/
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 from datetime import timedelta
+from dealer.git import git
 import os
 
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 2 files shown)


wetneb:
Awesome, thank you so much for tackling this! Just checking: you do have access to Sentry for dissem.in, right? It should just be a matter of logging in via GitHub as you are already a member of the organization, I think.

beckstefan:
Yes, I can log in and I see issues, so that works.

But I cannot link the repo and these things as in the link in #608 . That's up to you.

wetneb:
Great, doing now then.

Phyks:
Great, thanks!

My two cents on this, it might be worth enclosing https://github.com/dissemin/dissemin/pull/714/files#diff-e40c39db1e52f87d2d17dfa13c2b4a78R63 in a `TypeError` catch to avoid issues when downloading a ZIP file for Dissemin for instance or not having a Git repo behind.

PR #715, opened 2019-04-15
HTML validity

Phyks:


Phyks:
Here are a few fixes to make the HTML valid (using W3C validator). HTML and CSS should be valid now (up to a minor things from libs on which we have no control).

I'm not 100% convinced the second commit is worth it, I'm open to discussion and suggestions here. I am using http://www.html-tidy.org/ and the idea was to have HTML validation in the tests, to avoid creating such errors. It seems however than HTML Tidy differs from the official W3C validator on various points (it does not create issues for some things the W3C validator does, and creates issues on things which are considered valid by the W3C validator).

Not sure what should be done here, perhaps it would actually be better to submit the resulting HTML content to the W3C validator API (https://github.com/validator/validator/wiki/Service-%C2%BB-Input-%C2%BB-POST-body)?

papers/tests/test_pages.py
```diff
@@ -64,6 +66,24 @@ def checkHtml(self, resp):
             print(resp.content)
             print("HTML validation error: "+str(e))
```
wetneb: How does your new validation code below compare with the one we had above? If it is stricter, then can we remove the previous checks above, and drop the dependency to html5lib?

Phyks: I think this is a superset of the checks made by `html5lib` indeed. `html5lib` is trying to parse the HTML and will fail on invalid (in the sense of not parsable) HTML.

My test will consider additional checks such as invalid attributes or this kind of thing.

However, I'm wondering whether we should not in the end stick with `html5lib` and use the W3C Validator API to check for full HTML validity, I'm not 100% confident `html-tidy` is doing everything right.

Phyks: Actually, the W3C validator can run locally (https://github.com/validator/validator)! I'm using it (actually the Python `html5validator` wrapper).

Bonus points, I found a couple of small display issues like this (such as the labels not working as expected in the upload form) :)

Phyks:
Should be ready, with automatic testing against W3C validator (running locally).

PR #718, opened 2019-04-15
Fix and document the /api/query endpoint.

### LINKED ISSUE #717: /api/query is not documented
https://github.com/dissemin/dissemin/blob/master/papers/api.py#L150

this should be documented in https://dissemin.readthedocs.io/en/latest/api.html

wetneb:
Closes #717.

### DIFF
--- a/doc/sphinx/api.rst
+++ b/doc/sphinx/api.rst
@@ -35,6 +35,11 @@ to https://dissem.in/api/query, as follows::
 The date field can contain coarser dates such as ``2009-07`` or ``2009``, and authors can also be specified
 as plain text with ``{"plain":"Anne Moyer"}`` instead of ``{"first":"Anne","last":"Moyer"}``.
 
+This API method uses the internal paper deduplication strategy in Dissemin to match the bibliographic
+reference to a known paper in the database. This deduplication is done by computing a unique key (called fingerprint)
+from the title, authors and publication date. Therefore, this API method will always return at most one paper, 
+unlike the search endpoint below which works like traditional search engines.
+
 Searching for papers
 --------------------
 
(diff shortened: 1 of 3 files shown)


papers/api.py
```diff
@@ -213,7 +213,11 @@ def api_paper_query(request):
     except ValueError:
         raise BadRequest('Invalid paper')
 
-    return {'status': 'ok', 'paper': p.json()}
+    try:
```
Phyks: The `BarePaper.create` call above is only used to compute fingerprint? Or is any side effect expected? If this is the case, it could probably use a comment :)

doc/sphinx/api.rst
```diff
@@ -14,6 +14,27 @@ You can retrieve Dissemin's metadata for a specific paper by DOI:
...
+---------------------------
+
+When the DOI or the Dissemin ID are not known, it is possible to retrieve a paper by title,
+authors and publication date. This is done by posting a JSON object encoding this medatata
+to https://dissem.in/api/query, as follows::
+
+    curl -H "Content-Type: application/json" -d '{"title":"Refining the Conceptualization of an Important Future-Oriented Self-Regulatory Behavior: Proactive Coping", "date":"2009-07-01","authors":[{"first":"Stephanie Jean","last":"Sohl"},{"first":"Anne","last":"Moyer"}]}' https://dissem.in/api/query
+
+The date field can contain coarser dates such as ``2009-07`` or ``2009``, and authors can also be specified
+as plain text with ``{"plain":"Anne Moyer"}`` instead of ``{"first":"Anne","last":"Moyer"}``.
```
Phyks: Maybe there should be a word about the fingerprinting strategy?

That is, from the API user perspective, how exact should this search be (and what is the difference here with a `/search` call for instance).

papers/api.py
wetneb: Done!

doc/sphinx/api.rst
wetneb: Done!

Phyks:
:+1:

Issue #716, opened 2019-04-15
Refetch ORCID publications is broken


Phyks:
Steps to reproduce:

1. Go to https://dissem.in/my-profile
2. Click "Refresh papers"
3. All papers are displayed, not only the ones from the researcher.

When clicking "refresh papers", the XHR call is bogus, it is https://dissem.in/ajax/search. This is probably linked to a recent change actually (such as fixing the JSON bug when going backwards).

a3nm:
I can reproduce this. Also, when selecting a high page number (such as 10), it leads to a 404.

Phyks:
Same issue happens on a researcher page as well, when searching for instance.

Issue #721, opened 2019-04-16
Dissemin requires access to your ORCID name


Phyks:
When no name is public in ORCID, we end up on a very ugly and unstyled page:

![Capture d’écran 2019-04-16 à 12 11 15](https://user-images.githubusercontent.com/3856586/56201417-1c924700-6041-11e9-83d1-f957c7c3f859.png)

It seems that the issue might be more profound as well. Steps to reproduce:

1. Create an account on Dissemin with an ORCID account without any public information.
2. Go to "My-profile", you end up with the unstyled page.
3. Change the visibility of your ORCID name.
4. Go back to Dissemin, same issue.

My sandbox profile is https://sandbox.orcid.org/0000-0002-7013-2493, I may have missed something.

wetneb:
I think we only refetch the ORCID profile if it has not been fetched recently - this should probably be bypassed if we haven't got any public name for the user.

a3nm:
Hmm, just a side comment but going to my profile showed again the page that says "Dissemin requires access to your ORCID name". For some reason though, logging out and logging in again fixed the issue.

Phyks:
> For some reason though, logging out and logging in again fixed the issue.

Let's keep an eye open on this one. Let me know if you or anyone near you experiences this screen again without reason. Thanks!

a3nm:
This being said, if the message can still get shown to users (for a valid reason this time), then I still think it should be significantly improved: explain that names in ORCID have a visibility, that Dissemin requires the name to be set to "public", and have a link plus a list of steps (ideally with screenshots) that the user can follow to fix the problem. Currently I think the message is essentially incomprehensible to people who are not already very familiar with how ORCID works. Do you see what I mean?

Phyks:
The message will likely be shown to someone with a private name in their ORCID profile. It should look like https://user-images.githubusercontent.com/3856586/57301306-ffe5ae00-70d8-11e9-96ae-57885b5b3638.png.

I agree, the wording should be clearer. Reopening this one to track the clarity issue of the message.

PR #722, opened 2019-04-16
Fix issue #716: refetch and search not working on researchers page

Phyks:
This should solve the issue #716 about search and refetch from ORCID broken.

Tested against researcher page. **Untested** against "my profile" page, due to #721 :/

### DIFF
--- a/papers/urls.py
+++ b/papers/urls.py
@@ -30,9 +30,6 @@ urlpatterns = [
     url(r'^$', views.index, name='index'),
     # Paper views
     url(r'^search/$', views.PaperSearchView.as_view(), name='search'),
-    # This AJAX view is a trick to avoid displaying JSON output in browser when
-    # going backwards, see #198.
-    url(r'^ajax/search$', views.PaperSearchView.as_view(), name='ajax-search'),
     url(r'^search/advanced$', views.AdvancedPaperSearchView.as_view(),
         name='advanced-search'),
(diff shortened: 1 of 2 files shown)


papers/urls.py
```diff
@@ -51,6 +48,13 @@
         views.PublisherPapersView.as_view(), name='publisher-papers'),
     url(r'^journal/(?P<journal>\d+)/$',
         views.JournalPapersView.as_view(), name='journal'),
+    # These AJAX views are a trick to avoid displaying JSON output in browser
+    # when going backwards, see #198.
+    url(r'^ajax/search$', views.PaperSearchView.as_view(), name='ajax-search'),
+    url(r'^ajax/r/(?P<researcher>\d+)/(?P<slug>[\w-]*)$', views.ResearcherView.as_view(),
+        name='ajax-researcher'),
+    url(r'^ajax/my-profile$', views.myProfileView,
+        name='ajax-my-profile'),
```
wetneb: I guess we should not need this URL, since `ajax-my-profile` is not used anywhere else, right? Given that the `my-profile` view just reuses the researcher view, it will have the `ajax-researcher` URL in its context, I think.

Phyks: Indeed, well catched, thanks!

Phyks:
Should be fine and working. However, a second look is probably nice to have on this one :)

wetneb:
Thanks a lot, deployed!

Issue #687, opened 2019-03-29
Improve static assets loading and caching


Phyks:
See for instance

https://developers.google.com/speed/pagespeed/insights/?hl=fr&url=dissem.in. 

https://developers.google.com/speed/pagespeed/insights/?hl=fr&url=https%3A%2F%2Fdissem.in%2Fsearch%2F%3Fauthors%3Dverney

https://developers.google.com/speed/pagespeed/insights/?hl=fr&url=https%3A%2F%2Fdissem.in%2Fr%2F4162%2Flucas-verney

This should have a net impact on mobile devices and search engines ranking.

beckstefan:
For all the JS (that Google is complaining about) we can use "defer", however this would mean to make a dissemin.js (defer that too) and collect all the spread JS into that in case it relies on some JS-library that is defered.

Phyks:
> For all the JS (that Google is complaining about) we can use "defer", however this would mean to make a dissemin.js (defer that too) and collect all the spread JS into that in case it relies on some JS-library that is defered.

Indeed, that's more or less what I was planning to do mid-term as part of #563. I'd like to investigate using a bit more modern JS stack (Yarn/NPM, Webpack) to build the JS code into single entries which could be minified and loaded more efficiently.

In this issue, I'm mainly focusing on the server side configuration so far (ensuring gzip, caching strategies and other HTTP headers).

beckstefan:
I've done the above test on sandbox with the new bootstrap, which loads less JavaScript. There is a significant improvement (around 70 points).

However, I do not use defer, which should give us some extra points.

Another complain from Google is indeed caching. As it seems, we currently do no caching of our static files at all (nevertheless, the browser uses its cache).

PR #733, opened 2019-04-29
Add an example Apache2 vhost

Phyks:
This is an attempt at fixing https://github.com/dissemin/dissemin/issues/687. This would also provide users with a typical vhost to start from, to help them get starting hosting a Dissemin instance.

So far,
- We don't use any security HTTP headers (content security policy etc) on the production instance.
- We have a [B grade](https://www.ssllabs.com/ssltest/analyze.html?d=dissem.in) SSL score, part of the issues are easily fixable (the rest being difficult without dropping support for older browsers)
- We lack gzip and caching strategy for the static assets, see #687.
- We don't have CORS headers for the API, therefore preventing from using it from within the browser.

Note that this is a work in progress and currently untested :) Definitely not ready for a merge yet.

Left to do:

- [ ] We should have `Content Security Policy` headers to state where the resources can be fetched from.
- [ ] We should add <s>gzip /</s> caching for the static assets. Gzip is done, caching policy should be kept for the future I think.

These are left as `TODOs` in the code and I think it will be better to handle them separately in a future PR (when refactoring and updating the JS dependencies for instance).

### DIFF
--- a/provisioning/apache2-vhost.conf
+++ b/provisioning/apache2-vhost.conf
@@ -6,6 +6,7 @@
     Redirect permanent / https://dissem.in/
 </VirtualHost>
 
+SSLStaplingCache        shmcb:/var/run/ocsp(128000)
 <VirtualHost *:443>
     ServerAdmin webmaster@localhost
     ServerName dissem.in
@@ -30,7 +31,6 @@
     SSLUseStapling          on
     SSLStaplingResponderTimeout 5
     SSLStaplingReturnResponderErrors off
-    SSLStaplingCache        shmcb:/var/run/ocsp(128000)
 
     # Security headers
     # ================
--- a/provisioning/apache2-vhost.conf
+++ b/provisioning/apache2-vhost.conf
@@ -2,17 +2,6 @@
     # Upgrade insecure connections to HTTPS.
     ServerAdmin webmaster@localhost
     ServerName dissem.in
-    ServerAlias prod.dissem.in www.dissem.in
-
-    Redirect permanent / https://dissem.in/
-</VirtualHost>
-
-<VirtualHost *:443>
-    # Create a VirtualHost to redirect any alternative (legacy) domain
-    # to "dissem.in" in a permanent way.
-    ServerAdmin webmaster@localhost
-    ServerName www.dissem.in
-    ServerAlias prod.dissem.in
 
     Redirect permanent / https://dissem.in/
 </VirtualHost>
--- a/provisioning/apache2-vhost.conf
+++ b/provisioning/apache2-vhost.conf
@@ -1,0 +1,119 @@
+<VirtualHost *:80>
+    # Upgrade insecure connections to HTTPS.
+    ServerAdmin webmaster@localhost
+    ServerName dissem.in
+    ServerAlias prod.dissem.in www.dissem.in
+
+    Redirect permanent / https://dissem.in/
+</VirtualHost>
+
+<VirtualHost *:443>
+    # Create a VirtualHost to redirect any alternative (legacy) domain
+    # to "dissem.in" in a permanent way.
+    ServerAdmin webmaster@localhost
+    ServerName www.dissem.in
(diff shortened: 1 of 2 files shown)


provisioning/apache2-vhost.conf
```diff
@@ -0,0 +1,83 @@
...
+
+    # SSL configuration
+    # =================
+    SSLEngine on
+    SSLCertificateFile    /etc/letsencrypt/live/dissem.in/cert.pem
+    SSLCertificateKeyFile /etc/letsencrypt/live/dissem.in/privkey.pem
+    SSLCertificateChainFile /etc/letsencrypt/live/dissem.in/fullchain.pem
+    # Use HSTS
+    # TODO: Include subdomains? Preload?
+    Header always set Strict-Transport-Security "max-age=15768000"
```
Phyks: This header states that `dissem.in` is available over HTTPS and tells the browser to remember this for 6 months. Therefore, next time the user is sent to the HTTP version (through a link for instance), the browser will automatically upgrade to HTTPS transparently, before making any HTTP call.

The downside is that we should always have a working SSL stack (or the site will not be browsable at all, one cannot overload HSTS policy errors). This should be fine with the current Let's Encrypt setup and a bit of monitoring.

Question here:
- [ ] Do we want to include subdomains? So that upon visiting `dissem.in`, any `*.dissem.in` will be accessed on HTTPS next time. I think so, but we should make sure we have HTTPS available on all subdomains. `dev.dissem.in` and `association.dissem.in` are both fine. Do we have any others?
- [ ] Do we want to include preloading? So that whenever a user visits our website, the rule is sent to a central list and all users will now make queries in HTTPS by default? I think so as well :)

a3nm: On my own web properties I've always been wary of enabling HSTS, because of the significant downside (broken SSL setup means that the website is completely unreachable for all users -- HSTS also mandates that browsers must not allow users to override errors). Broken HTTPS (e.g., on subdomains) has happened in the past, so that's a genuine risk. My understanding of HSTS is also that this choice is "irreversible" in the sense that if we change our mind about this, we will be stuck with clients expecting HTTPS to work until expiration of the max-age value.

The upside is that user traffic is encrypted, but it's not like I see any real benefit to do this for Dissemin traffic.

I don't know, though: as Let's Encrypt has been around for a long time now and HTTPS is mostly working fine, maybe I'm just being too risk-averse here.

Phyks: That's precisely the main issue with having HSTS (it being irreversible etc). For the record, the current SSL configuration is based on Mozilla's recommendation (https://mozilla.github.io/server-side-tls/ssl-config-generator/) and the oldest browsers supported by this configuration probably already see a broken Dissemin interface due to lacking some modern CSS / JS abilities.

My opinion is that for a website such as Dissemin aiming at a broad audience, HTTPS and HSTS should be implemented. Then, this implies we should have certificates monitoring and a proper SSL setup (but we should anyways have such monitoring in my opinion if we target broad audience :).

Phyks: Checking with the recent DNS entries update, I found doai and the server hostnames which do not have a valid SSL cert. The rest should be fine.

My proposal would be to keep

```
Header always set Strict-Transport-Security "max-age=15768000; preload"
```

This means that upon visiting `dissem.in`, the info that HTTPS is available and should be used will be kept for 15768000 seconds (6 months) and this info will be sent to the HSTS preload list (built into most recent browsers).

I would not add `includeSubDomains` as we might miss HTTPS on subdomains or not monitor them correctly.

What do you think?

Phyks:
Now deployed on https://dissem.in/. Let me know if you find anything suspicious.

- SSL grade is now A+, https://www.ssllabs.com/ssltest/analyze.html?d=dissem.in.
- Lighthouse score is still average, mainly because of the lack of caching policy for static assets, https://developers.google.com/speed/pagespeed/insights/?url=dissem.in&tab=mobile.

PR #736, opened 2019-04-30
WIP: Content for Startpage

beckstefan:


startpage_content.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Green Open Access
+_(Citation advantage of OA https://sparceurope.org/what-we-do/open-access/sparc-europe-open-access-resources/open-access-citation-advantage-service-oaca/ )_
+
+Many researchers do not use their right to make their papers freely available online, in addition to the paywalled version offered by traditional publishers.
+
+There are two major advantages of unpaywalling your papers:
+1. There are a lot of studies giving evidence, that freely available publications have a higher citation rate. 
+This is called the _Open Access Citation Advantage_. You can find a list of studies on https://sparceurope.org/what-we-do/open-access/sparc-europe-open-access-resources/open-access-citation-advantage-service-oaca/oaca-list/ 
+As a conclusion: Your research is not only more acknowledged, but pushes scientific advantage in general, because your research better spreads!
```
wetneb: I would first start with something general about articles being available to readers (peers / patients / the industry / the general public) and only mention citation rates in a second sentence (because citation rates are debatable metrics, researchers should not try to optimize them per se - they should rather try to increase the reach of their publications in a general sense).

startpage_content.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Success-Story
+_(Get some data from zenodo, maybe contact the depositor?)_
+
+The publication __%publication__ was uploaded on __%date__ and got __%number__ downloads and __%number__ views on Zenodo!
+
+The depositor says: Uploading this publication on dissem.in was the best idea I ever had!
+
+We have another __%number__ publications with at least __%number__downloads on Zenodo.
+
```
wetneb: We would need to find good examples of this so that it is really convincing - I suspect the number of views / downloads on Zenodo might not be extremely appealing by themselves…

This section could potentially be replaced / complemented by the list of recent deposits.

beckstefan: Yes. First we would need to grab data, evaluate it, to make sure that we have adorable numbers and then write some emails hoping for good answers. Of course, the results may lead to a dead end.

startpage_content.md
```diff
@@ -0,0 +1,62 @@
+This WIP PR is for preparing content for the start page.
+
+Please commit your ideas until 2019-05-11 so that we can discuss on the next web meeting about the design.
+
+Solutons could be:
+
+* screens (can take a lot of content)
```
a3nm: Not sure what you mean by "screens"... Screenshots?

startpage_content.md
a3nm: +1 Antonin for citations being only a secondary point. Several more comments:

- We need to have as persistent content a description of what the service proposes, before giving context about green OA. We currently have " Dissemin detects papers behind pay-walls and invites their authors to upload them in one click to an open repository." I think that's OK except we may want to add some term like "green OA" to this if we have more detailed explanations afterwards.
- I think the text should be shorter. I'd keep the current "Many researchers do not use their right to make their papers freely available online, in addition to the paywalled version offered by traditional publishers." and then follow by a list of advantages in a few words:
  - Ensure that libraries don't need to pay overpriced subscriptions to scientific journals.
  - Make sure that everyone can read your research, including people outside of established research labs
  - Give more impact to your research (just link to the relevant study)

startpage_content.md
```diff
@@ -0,0 +1,62 @@
...
+Many researchers do not use their right to make their papers freely available online, in addition to the paywalled version offered by traditional publishers.
+
+There are two major advantages of unpaywalling your papers:
+1. There are a lot of studies giving evidence, that freely available publications have a higher citation rate. 
+This is called the _Open Access Citation Advantage_. You can find a list of studies on https://sparceurope.org/what-we-do/open-access/sparc-europe-open-access-resources/open-access-citation-advantage-service-oaca/oaca-list/ 
+As a conclusion: Your research is not only more acknowledged, but pushes scientific advantage in general, because your research better spreads!
+1. Libraries do not need to by overpriced electronic subscriptions to journals, when they can afford them at all.
+
+## Open repositories
+Uploading your papers on your own webpage is not enough. Such copies are less stable and harder to find than documents uploaded to well-indexed repositories.
```
a3nm: I think this sentence should be near the end as it's a reply to a common objection but not the most important point.

startpage_content.md
```diff
@@ -0,0 +1,62 @@
...
+There are for a lot of major repositories like arXiv, chemRxiv and institutional repositories. Nearly all of them are searched by BASE and CORE. And they ensure a long-term availability of your research.
+
+Dissemin searches for copies of your papers in a large collection of open repositories and tells you which ones cannot be accessed.
```
a3nm: I'd rather say "... which ones it has not been able to access." In my opinion this should be the first sentence.

Then could follow something like: "Our repositories are those indexed by CORE and BASE" (with links to these services), including arXiv, chemRxiv, and institutional repositories such as HAL and ...".

And only then do we explain that we don't take personal web pages into account. (Maybe we could also say something against academic social networks?)

startpage_content.md
```diff
@@ -0,0 +1,62 @@
...
+## Open repositories
+Uploading your papers on your own webpage is not enough. Such copies are less stable and harder to find than documents uploaded to well-indexed repositories.
+There are for a lot of major repositories like arXiv, chemRxiv and institutional repositories. Nearly all of them are searched by BASE and CORE. And they ensure a long-term availability of your research.
+
+Dissemin searches for copies of your papers in a large collection of open repositories and tells you which ones cannot be accessed. 
+
+## Statistics
+_(With d3js it sould be easy to create some fancy, dynamic on-the-fly-graphs)_
+
+Since __%date__ __%number__ of publications have been uploaded via dissemin. This make __%number__ publications per week.
```
a3nm: Also have stats about how many pubs were *accessed* via Dissemin. This would be counted as the number of clicks to a repository via Dissemin -- but requires us to add some JS or change our links to detect clicks to external websites.

startpage_content.md
```diff
@@ -0,0 +1,62 @@
...
+_(Get some data from zenodo, maybe contact the depositor?)_
+
+The publication __%publication__ was uploaded on __%date__ and got __%number__ downloads and __%number__ views on Zenodo!
+
+The depositor says: Uploading this publication on dissem.in was the best idea I ever had!
+
+We have another __%number__ publications with at least __%number__downloads on Zenodo.
+
+
+## Dissemin for libraries
```
a3nm: I doubt that libraries are a large proportion of our visitors, so I think this should maybe be a separate page. There would be something like "are you a library? Discover here how you can use Dissemin" with a link.

I also think the text should be clearer, in particular separating what we offer ("you can use...") vs what we ask ("you can advertize..." should probably be "advertise" btw). "connect your repository" should be explained better and is arguably targeted towards other people, namely, repository admins.

Maybe we should have a list of links for different audiences, like:

- Are you a researcher? Discover how to start depositing your papers with Dissemin
- Are you a librarian? (or "managing a lab"?) Discover how to use Dissemin to find the papers from your institution
- Are you a repository administrator? Discover how to index your repository in Dissemin
- Are you a developer? Discover how to use our API

startpage_content.md
Phyks: Screens in the sense of what Unpaywall (https://unpaywall.org/) is doing for instance (although this is not the most screen-esque website). The homepage is made of "slides" stacked vertically. The first one occupies the full screen.

It's mostly slides and a matter of organizing them (either in a carousel, with auto-sliding, or stacked vertically). :)

Phyks:
I set up a Framadocument so we can collaboratively edit this one https://annuel.framapad.org/p/Dissemin-frontpage (see minutes of the developer meeting to come).

Issue #739, opened 2019-05-02
"Paper object" in the deposits page


a3nm:
In the "My deposits" page for logged in users, papers are presented as "Paper object (xxx)" with xxx being some number. This should instead be the title of the paper.

Phyks:
Could you take a screenshot please? (shame on me, I don't have any deposit through Dissemin)

a3nm:
![https://a3nm.net/share/2019_05_02_744855.png](https://a3nm.net/share/2019_05_02_744855.png)

a3nm:
Yup that's fine for me now, thanks!

PR #742, opened 2019-05-02
Show paper title instead of Paper raw object in My deposits page.

Phyks:


Phyks:
Not easy to check, but I *think* this should fix https://github.com/dissemin/dissemin/issues/739. cc @a3nm

PR #748, opened 2019-05-02
Dissemin non-bibliographic metadata schema

beckstefan:
A version for the metadata schema and documentation that was mentioned in the web conference.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,6 @@ runfcgi.pid
 
 # build files for doc
 doc/sphinx/_build/
-doc/sphinx/*.rst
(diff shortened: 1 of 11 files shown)


beckstefan:
> The use of `:download:` in the Sphinx docs is clever - I assume it is not necessary for instances to serve the XSD directly then.

Dunno. The main purpose is to make it available for users (repo admins) as documentation for their own implementation, not for validation. For the latter purpose serving the file as static under the namespace URI seems more legit: RTD guarantees persistent documentation, the documentation itself might change and thus the file URL.

Yet, we do not need to decide this right now.

PR #750, opened 2019-05-02
Indicating the status of HAL affiliations, fixes #741

### LINKED ISSUE #741: Indicating the status of HAL affiliations
When indicating an affiliation for HAL, in the search field with autocompletion, all matching affiliations are listed, but not all of them are good. In the response of the AJAX call, these affiliations are indicated either as "valid" (i.e., preferred), "incoming" (i.e., have not been validated yet), or "old" (i.e., should no longer be used).

It would probably be better to group the affiliations (probably in an optgroup), first the "valid" ones ("Current institutions" / "Institutions existantes"), then the "incoming ones" ("Unverified institutions" / "Institutions proposées"), and then the "old" ones ("Obsolete institutions" / "Institutions obsolètes").

PierreSenellart:


wetneb:
As a planning step for institutional integration, I have proposed this property in Wikidata:
https://www.wikidata.org/wiki/Wikidata:Property_proposal/HAL_institution_ID
This should help us derive the HAL affiliation from the ORCID employer, for instance.

PierreSenellart:
OLD records might still be useful for historical papers (to upload a paper with a historical affiliation that does not exist any more).

PR #752, opened 2019-05-07
Dropdown for Repo-Protocol

beckstefan:
This adds a dropdown without changing any logic of registration.

Do we need tests for this? I am not even sure, how to perform them after all.

### DIFF
--- a/deposit/forms.py
+++ b/deposit/forms.py
@@ -22,6 +22,7 @@
 
 from deposit.models import Repository
 from deposit.models import UserPreferences
+from deposit.registry import protocol_registry
 from django import forms
 from django.urls import reverse_lazy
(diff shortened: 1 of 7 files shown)


deposit/forms.py
```diff
@@ -139,5 +140,22 @@ def __init__(self, *args, **kwargs):
             Submit('submit', _('Save')),
         )
 
-
-
+class RepositoryAdminForm(forms.ModelForm):
+    """
+    Changes widget of the protocol field to a select (dropdown) of AdminView of Repository. This way no free has to be entered for the protocol to choose.
```
Phyks: free "text" ? (I think there is a missing word here)

beckstefan: Yep, inserted, thanks!

PR #754, opened 2019-05-09
Fix ORCID ingestion

### LINKED ISSUE #723: No reason set for skipped works in ORCID importer
While importing the ORCID dump:
```
2019-04-15 11:32:00 INFO dissemin.papers.orcid:138  https://pub.orcid.org/v2.1/0000-0002-9005-9502/works
2019-04-15 11:32:01 INFO dissemin.papers.orcid:138  https://pub.orcid.org/v2.1/0000-0002-9005-9502/works/5101707,5101717,5101709,5101713,5101712,5101711,5101714,5101716,5101715,5101706,5101708,5101710
Traceback (most recent call last):
  File "/usr/lib/python3.5/code.py", line 91, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
  File "/home/dissemin/prod/ingest_orcid.py", line 13, in <module>
    o.bulk_import('/home/pintoch/summaries')
  File "/home/dissemin/prod/backend/orcid.py", line 287, in bulk_import
  File "/home/dissemin/prod/backend/orcid.py", line 220, in fetch_orcid_records
    continue
AttributeError: 'OrcidWork' object has no attribute 'reason'
```

Phyks:
Fix 'OrcidWork' object has no attribute 'reason' in ORCID dump
ingestion, fix #723.

This is tested and make the ORCID dump ingestion more robust. I added a bit of logging to have a better output and feedback on the progress.

I am currently running it on tulpan in a screen session.


wetneb:
To report progress and speed, I have made a generator wrapper that you could use for that:
https://github.com/dissemin/dissemin/blob/96b83f8332fcde1b7ff89eb49b47cb336055e226/backend/utils.py#L120

The idea is that you can use it with any generator - it returns the same generator, with the additional reporting logic.

backend/orcid.py
```diff
@@ -293,7 +293,8 @@ def bulk_import(self, directory, fetch_papers=True, use_doi=False,
                         if fetch_papers:
                             papers = self.fetch_orcid_records(
                                 orcid,
-                                profile=OrcidProfile(json=profile),
+                                profile=OrcidProfile(orcid_id=orcid,
```
Phyks: Is it a valid use case to actually create an `OrcidProfile` without any `orcid_id`? If this is not the case, I would actually update the `OrcidProfile` constructor to avoid this kind of bugs to appear again.

backend/orcid.py
```diff
@@ -244,37 +248,66 @@ def fetch_and_save(self, researcher, profile=None):
...
+                # Eventually skip first entries
+                if fname == start_from:
+                    seen = True
+                if start_from and not seen:
+                    continue
 
-                with open(path.join(root, fname), 'r') as f:
+                # Log progression
+                nb_profiles += 1
+                if nb_profiles % 1000 == 0:
```
Phyks: Should use https://github.com/dissemin/dissemin/blob/96b83f8332fcde1b7ff89eb49b47cb336055e226/backend/utils.py#L120 instead.

backend/orcid.py
wetneb: I think the intention that the constructor would indeed retrieve the ORCID id from the json representation if provided.

Phyks:
The latest commit uses the ORCID API dump to fill in `work-summaries` and then avoid one API call for empty profiles.

The remaining issue is for the works themselves. We don't use [DOIs and Crossref](https://github.com/dissemin/dissemin/blob/master/backend/orcid.py#L197-L205) when importing the ORCID dump, therefore we make a bunch of API calls to ORCID to fetch the [works](https://github.com/dissemin/dissemin/blob/master/backend/orcid.py#L208) for each profile. This is not included in the JSON conversion of the API dump.

This however is available in a separate dump, in XML format. Maybe our best bet would be to actually preprocess this XML dump and convert it to JSON format and do the whole import process offline.

I don't have numbers with querying CrossRef API (as it is currently disabled for ORCID ingestion), but with all the queries to ORCID to fetch the works, it takes about 1500s to ingest 1k profiles, which is way too long.

nemobis:
Is it useful/expected to create an ORCID profile without any content other than the ORCID ID? https://dissem.in/r/187946/ursula-ellinghaus

Phyks:
Not sure about the usefulness of having empty profiles. This being said, the very costly part here is requests made to the ORCID API (to fetch works). This does not happen for empty profiles, which are already created without making any HTTP request with this PR, so they basically take milliseconds to be created.

Given the number of ORCID profiles to process, these milliseconds might matter. But anyways, 25M ORCID profiles (half of them) with the current setup would take more than 400 days to ingest :/

beckstefan:
> This however is available in a separate dump, in XML format. Maybe our best bet would be to actually preprocess this XML dump and convert it to JSON format and do the whole import process offline.

Maybe I am to naive here. Importing from this once and then query ORCID API on demand if necessary? Maybe with a "last updated" field in our DB? 
What exactly makes the ORCID API import that slow? 1.5 seconds per profile is a lot.

wetneb:
@beckstefan the ORCID API used to return an entire profile in one HTTP request. That is no longer the case: you need to retrieve publications by batches from them.

Yes it would make sense to import from the dump - it's just annoying that only XML format (and not JSON) is available now (so it means duplicating our parsing code for XML).

Phyks:
> What exactly makes the ORCID API import that slow? 1.5 seconds per profile is a lot.

Additionally to @wetneb's answer, the ORCID API dump provides everything in XML (whereas the live API we use is JSON). Only part of it (which is profile summaries) is usable with theconversion lib https://github.com/ORCID/orcid-conversion-lib/ to convert it to JSON. Therefore, we use the local dump converted to JSON for the first steps of the import (base profile infos, work summaries) and then we have to query the live API for batches of works. This is the slow part (and not easily runnable in parallel because of API rate limitations anyways), we should use the dump altogether.

> Yes it would make sense to import from the dump - it's just annoying that only XML format (and not JSON) is available now (so it means duplicating our parsing code for XML).

Or preprocessing the dump. Given the size of the dump and the amount of useless stuff for us so far (education info, employment infos etc), I'd rather have a pre-processing chain for extracting/filtering/JSONizing the XML dump.

wetneb:
> Or preprocessing the dump. Given the size of the dump and the amount of useless stuff for us so far (education info, employment infos etc), I'd rather have a pre-processing chain for extracting/filtering/JSONizing the XML dump.

Yeah, and the easiest way to do this might be to extend their utility actually… :-/

Phyks:
Note to self: it is not really practically doable to expand the full ORCID activities dump.

There are 1100 folders in the summaries dump:
```
$ ls /home/…/summaries | wc -l
1100
```

Expanding the ORCID API (filtering on the works only and skipping education, employment etc, one log line every 10k files extracted, stopped before finishing):
```
$ tar zxvf /home/…/orcidapi2.0_activities_xml_10_2018.tar.gz --wildcards activities/*/*/works/ | awk 'NR == 1 || NR % 10000 == 0'
$ ls activities/| wc -l
120
$ du -h --summarize activities
72GB
```

The whole ORCID API dump is then expected to be about 720GB large. The best approach would be to process folders by folder I think:

1. Iterate on the folders in the ORCID API summary dump.
2. For each folder, extract the corresponding activities, pre-process them (convert them from XML to JSON).
3. Then, ingest the folder in Python, off line. Go on with next folder.

I'll start by writing some quick and dirty script to convert the XML files to JSON.

Phyks:
So, I found a trick to overcome the tar gz limitations :) One of the main issues was that the top level directories (both dumps are of the form `TOP_LEVEL/<ORCID_ID>` with 1100 top level directories) were weirdly sorted in the activities (works) dump, which I cannot extract. I iterated over this order and matched it with the summaries, which should be totally fine now.

Current run time is about 10s to parse a top level directory (get a list of members) in Python, and then about 30s to extract the whole top level directory. We cannot reduce this much. Once this step is done, all the import is done offline, without any query to ORCID, which should be doable at maximum speed.

Here is a trace of the import running:

```

**Comment by Phyks** [comment]
> How about adding a quick test for the bulk_import method?

Sure, will have a look.

File: backend/orcid.py [hunk 3]
```diff
@@ -197,15 +214,21 @@ def fetch_orcid_records(self, orcid_identifier, profile=None, use_doi=True):
...
+            for idx, (success, paper_or_metadata) in enumerate(metadata):
                 if success:
-                    yield paper_or_metadata # We know that this is a paper
+                    yield paper_or_metadata  # We know that this is a paper
                 else:
                     put_codes.append(dois_and_putcodes[idx][1])
 
         # 2nd attempt with ORCID's own crappy metadata
-        works = profile.fetch_works(put_codes)
+        works = profile.fetch_works(
```
Phyks: Question rising from developper meeting: When we ingest the ORCID dump, we want to do it offline so we don't take into account the DOIs and the whole previous lines are skipped. We only use ORCID metadata (no DOI) which are incomplete.

So, ORCID dump should probably be imported **after** Crossref import is done and up to date. We could lookup for papers by DOIs (without queries to Crossref) if we have DOIs as well.

Any idea on the strategy here?

backend/orcid.py
```diff
@@ -244,37 +270,130 @@ def fetch_and_save(self, researcher, profile=None):
...
+                        members=members_to_extract
+                    )
+
+                    # Import the ORCID profiles for this toplevel directory
+                    # from the summaries dump
+                    summaries_files_with_speed = with_speed_report(
+                        os.listdir(summaries_path),
+                        name='ORCID profiles'
+                    )
+                    for summary_file in summaries_files_with_speed:
```
Phyks: Would it make sense to have parallel `for` loop here?

backend/management/commands/ingest_orcid.py
```diff
@@ -0,0 +1,13 @@
+from django.core.management.base import BaseCommand
```
Phyks: Same should be done for BASE import, see https://github.com/dissemin/dissemin/pull/760#discussion_r284234579.

doc/sphinx/dataingestion.rst
```diff
@@ -0,0 +1,28 @@
...
+
+Both are a tar gz archive of XML files. The dump of summaries should be
+extracted and converted to JSON using `the orcid conversion library
+<https://github.com/ORCID/orcid-conversion-lib/>`_. Let us assume the resulting
+JSON files are stored in ``/home/orcid/summaries``. The dump of activities
+should be kept as a tar gz archive, let us assume it is sitting at
+``/home/orcid/activities.tar.gz``.
+
+The dumps can then be imported using ``python manage.py ingest_orcid
+/home/orcid/activities.tar.gz /home/orcid/summaries``.
```
Phyks: Add doc about BASE dump import, see https://github.com/dissemin/dissemin/pull/760#discussion_r284234579.

backend/orcid.py
wetneb: I feel like we need to rethink the general architecture to handle this in a satisfactory way… If we can rely on the fact that Crossref is already imported, then ORCID should not be fetching its metadata a second time from the API, but only adding the ORCID information to the authors of the papers (adding an OaiRecord to keep track of the provenance).

But that means going away from the architecture where each paper importer emits a pure Python object (not linked to the database) representing all the paper metadata, and then a generic backend code deals with the saving to the DB. We currently don't follow that infrastructure completely, but that is the general direction I have been pushing towards.

Brainstorming session required!

nemobis:
> 25M ORCID profiles (half of them) with the current setup would take more than 400 days to ingest :/

Alternative view: if we had started 6 months ago, we'd be half done!

PR #757, opened 2019-05-12
Optimization for speed in Crossref import

### LINKED ISSUE #756: Index on publishers tables not used
We use `db_index=True` on Journal.name and Publisher.name but the index that creates is not used as we need an index on `UPPER(journal.name)` instead. We need to add a migration that creates these indices.

wetneb:
This adds various optimizations for speed to the paper importing code.

Closes #756.


wetneb:
Tested:
```
EXPLAIN SELECT * FROM papers_publisher WHERE UPPER(name) = UPPER('HFSP Publishing');
                                               QUERY PLAN                                               
--------------------------------------------------------------------------------------------------------
 Index Scan using papers_publisher_name_upper on papers_publisher  (cost=0.28..15.54 rows=15 width=131)
   Index Cond: (upper((name)::text) = 'HFSP PUBLISHING'::text)
(2 rows)

```

PR #760, opened 2019-05-13
BASE dump import and optimizations for speed

wetneb:
Apparently we didn't even have some code to import a dump, not sure exactly where this went… This is some incremental progress towards being able to stay in sync more efficiently with BASE.


wetneb:
When importing papers that are already there and do not need updating, this code runs at ~ 600 papers/sec in the sandbox. I think we should start running this in production to import the missing papers.

backend/oai.py
```diff
@@ -111,6 +114,43 @@ def ingest(self, from_date=None, metadataPrefix='oai_dc',
...
+        Given a path to a directory, representing an un-tar'ed BASE dump,
+        read all the bz2'ed files in it as a generator of OAI records
+
+        :param directory_path: the path to the directory where the BASE dump was un-tar'ed
+        :param metadata_prefix: the metadata prefix to read the records
+        """
+        filenames = os.listdir(directory_path)
+        namespaces = self.client.getNamespaces()
+        metadata_registry = self.client.getMetadataRegistry()
+        for filename in filenames:
```
Phyks: Question here (same could apply to ORCID import actually): Wouldn't it be possible to replace the `for` loops on filenames by `parfor` loops (https://joblib.readthedocs.io/en/latest/parallel.html) to iterate over the files in parallel. Looks feasible (as files are independent), not sure about the expected processing gain here?

Phyks: Actually, we are already using a bit of parallel processing downstream in the import. That's not the case for ORCID however, I think.

backend/oai.py
```diff
@@ -157,33 +204,47 @@ def process_record(self, header, metadata, format):
...
-                    processed_since_report = 0
-                    last_report = datetime.now()
+        with ParallelGenerator(listRecords, max_lookahead=max_lookahead) as g:
+            for record_group in group_by_batches(with_speed_report(g, name='OAI papers')):
+                oai_ids_to_last_updated = {
+                    record[0].identifier():record[0].datestamp() for record in record_group
+                }
+
+                # Fetch the last modified date of all these records in the DB
+                last_modified_in_db = { id : last_modified
```
Phyks: Nit: I would rather write it as

```python
last_modified_in_db = {
    ident: last_modified
    for ident, last_modified in OaiRecord.objects.filter(
        identifier__in=oai_ids_to_last_updated.keys()
    ).values_list('identifier', 'last_update')
}
```

(PEP8 + `id` is already a Python keyword)

backend/oai.py
```diff
@@ -157,33 +204,47 @@ def process_record(self, header, metadata, format):
...
+                               if record[0].identifier() in ids_to_update]
+
+                for paper in bare_papers:
+                    if paper is not None:
+                        try:
+                            with transaction.atomic():
+                                saved = Paper.from_bare(paper)
+                            return saved
+                        except ValueError:
+                            logger.exception("Ignoring invalid paper with identifier " + paper.oairecords[0].identifier)
```
Phyks: Better to use

```python
logger.exception("Ignoring invalid paper with identifier ", paper.oairecords[0].identifier)
```

(comma parameters for string formatting of `logging`).

backend/oai.py
```diff
@@ -157,33 +204,47 @@ def process_record(self, header, metadata, format):
...
+
+                # Fetch the last modified date of all these records in the DB
+                last_modified_in_db = { id : last_modified
+                    for id, last_modified in
+                    OaiRecord.objects.filter(identifier__in=oai_ids_to_last_updated.keys()).values_list('identifier', 'last_update')
+                }
+
+                # Deduce the list of papers to update
+                ids_to_update = set()
+                for id, last_updated_in_base in oai_ids_to_last_updated.items():
```
Phyks: Same comment about `id` being a Python keyword. Not a big deal, but this hides it.

backend/tests/test_oai.py
```diff
@@ -65,6 +65,29 @@ def delete(self, identifier):
...
+        titles = [record.getField('title') for _, record, _ in records]
+        self.assertEqual(len(records), 20)
+        self.assertTrue(['Modularizing the Elimination of r=0 in Kleene Algebra'] in titles)
+        
+    def test_load_dump(self):
+        """
+        Loads up a dump in the DB.
+        """
+        self.base_oai.load_base_dump(os.path.join(self.testdir, 'data/example_base_dump'))
+        # Loading it a second time should be very quick
```
Phyks: "Very quick"? Can we have a metrics here? Typically, if second load is expected to be about one order of magnitude faster, we could have an extra test on timing here, to assert this.

```python
start = time.time()
self.base_oai.load_base_dump(os.path.join(self.testdir, 'data/example_base_dump'))
duration = time.time() - start
start = time.time()
self.base_oai.load_base_dump(os.path.join(self.testdir, 'data/example_base_dump'))
assert((time.time() - start) < duration / 10)
```

backend/oai.py
```diff
@@ -157,33 +204,47 @@ def process_record(self, header, metadata, format):
...
+                    if id not in last_modified_in_db or last_modified_in_db[id].date() <= last_updated_in_base.date():
+                        ids_to_update.add(id)
```
Phyks: Is this `for` loop large? Might be (a bit) more efficient to do it with a generator on the fly:

```python
ids_to_update = set(
    ident
    for ident, last_updated_in_base in oai_ids_to_last_updated.items() 
    if ident not in last_modified_in_db or last_modified_in_db[ident].date() <= last_updated_in_base.date()
)
```

backend/translators.py
```diff
@@ -147,7 +147,7 @@ def translate(self, header, metadata):
             self.add_oai_record(header, metadata, paper)
             return paper
         except ValueError as e:
-            logger.warning("OAI record "+header.identifier()+" skipped:\n", e, exc_info=True)
+            logger.warning("OAI record "+header.identifier()+" skipped:\n%s", e, exc_info=True)
```
Phyks:
```python
logger.warning("OAI record %s skipped:\n%s", header.identifier(), e, exc_info=True)
```

backend/utils.py
```diff
@@ -151,3 +151,17 @@ def wrapped_generator(*args, **kwargs):
...
+    """
+    Given a generator, returns a generator of groups of at most batch_size elements.
+    """
+    current_batch = []
+    for item in generator:
+        current_batch.append(item)
+        if len(current_batch) == batch_size:
+            yield current_batch
+            current_batch = []
+    if len(current_batch) > 0:
```
Phyks: `if current_batch` (no `len` needed)

backend/oai.py
```diff
@@ -111,6 +114,43 @@ def ingest(self, from_date=None, metadataPrefix='oai_dc',
         records = self.client.listRecords(**args)
         self.process_records(records, metadataPrefix)
 
+    def load_base_dump(self, directory_path):
```
Phyks: We could probably expose this function in a `manage.py` command, similar to https://github.com/dissemin/dissemin/pull/754/files#diff-3e67c0e4d4dd1d92d9ae8a339953cbdbR1.

I can make it and complete the doc introduced as part of #754 with BASE counterpart in #754.

backend/oai.py
wetneb: I am not sure about reading the files in parallel - I would assume DB access remains the bottleneck here, but I haven't actually checked that. It's true that decompressing and parsing large XML files has a cost.

backend/oai.py
wetneb: Great, that would be very useful I think!

backend/tests/test_oai.py
wetneb: Yes, why not, but I try to avoid timing tests as they are not very reliable… Travis is sometimes slower at certain times, so this is prone to generate spurious build failures, which are annoying. I just tried with a factor of 5 and it failed the first time (running Pytest from the command line) but succeeded the second (from Eclipse).

Phyks: I was thinking that relative timings would be more robust than absolute timings, but apparently it's not so convenient. :+1: for not having build failures due to unreliable CI, let's skip this.

backend/oai.py
Phyks: Cool, I moved this to #754.

backend/oai.py
beckstefan: At the moment I expect the cost of implementation higher as the cost of not using `parfor`, so I leave as it is.

beckstefan:
The tests are passing.

In the OaiRecords from BASE we have irregular pubtypes `article-journal`, `''` and `None`.

The last one seems to come from the fact that on production the value is not set, but the to other seem mysterious to me at the moment. I'd like to investigate this, then we can probably merge and correct our incorrect types.

PR #762, opened 2019-05-14
Deposit Logging

beckstefan:
As discussed in the web conference we add

- [x] Lookup of DepositRecords by Paper id
- [x] Logging to Sentry in case of failure

### DIFF
--- a/deposit/admin.py
+++ b/deposit/admin.py
@@ -30,6 +30,7 @@ class DepositRecordAdmin(admin.ModelAdmin):
     list_display = ('identifier', 'paper', 'user')
     list_filter = ['repository']
     raw_id_fields = ('paper', 'user', 'oairecord')
+    readonly_fields = ('date', )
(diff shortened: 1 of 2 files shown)


deposit/admin.py
```diff
@@ -28,7 +28,9 @@
 
 class DepositRecordAdmin(admin.ModelAdmin):
     list_display = ('identifier', 'paper', 'user')
+    list_filter = ['repository']
     raw_id_fields = ('paper', 'user', 'oairecord')
+    search_fields = ('paper__pk', 'paper__title')
```
beckstefan: This adds a simple search field above the list of DepositRecords. Is this fine enough @wetneb ? Showing the DepositRecordIDs in a PaperView is non-trivial (but doable, if needed).

wetneb: That's great, thanks a lot!

PR #763, opened 2019-05-15
Issue 761 Model Licenses

### LINKED ISSUE #761: Model for Licences
Licenses are currently hard-coded in the source.

Move them to a model with option for translation with https://pypi.org/project/django-modeltranslation/

Related to #364

beckstefan:
Closes #761 

Adds a License model. LicenseChooser model is used to bring together licenses and repositories. LicenseChooser is displayed inline in repository page on admin site.

This PR has migrations and translations.

- [x] Migration altering non-related fields (see below)
- [x] Migration + predefined licenses ~~+ migrationtests~~
- [x] Admin-UI-Check
- [x] Tests
- [x] Repository class + tests
- [x] Forms + tests for existing protocols/repositories
  - [x] ~~HAL~~ (does not use licenses)
  - [x] Zenodo
  - [x] OSF (necessary?)
- [x] Documentation
- [x] Vinaigrette


beckstefan:
Running migrations yields

    deposit/migrations/0011_auto_20190515_1045.py
      - Create model License
      - Alter field oairecord on depositrecord
      - Alter field last_repository on userpreferences
      - Alter field preferred_repository on userpreferences
      - Add field licenses to repository

Is there a migration missing, that django wants to alter fields? Is this something to ignore and I just take it into my migration?

wetneb:
@beckstefan Oops sorry it does look like we are missing migrations indeed. Let me push them to master directly, so that you can rebase and only have relevant migrations in your PR.

beckstefan:
Thank you! :-)

deposit/migrations/0013_populate_licenses.py
```diff
@@ -0,0 +1,110 @@
+# Generated by Django 2.2.1 on 2019-05-15 14:53
```
beckstefan: @wetneb
I'm not sure, if we should keep this migration. For the migration on the production server this is rather nice, since it automatically predefines licenses and connects them with the current repos.

For tests this is not so good, because it adds initial data (that is probably not interfering, cause the licenses are just in the DB and repository entries are not changed since they do not exist).

I think best would be, to remove this migration from PR and manually execute the code via django shell after the migration.

Are you fine with that?

wetneb: I am not sure I fully understand the constraints you have, but my gut feeling is that it is totally fine to add initial data in migrations. What would be the downside of that?

I think we should try to make the deposit interfaces work out of the box as much as possible, and migrations are good at doing that. What I would propose is to add these licenses as migrations in the sub-apps for each protocol (`deposit.zenodo` and `deposit.osf`). This way, the licenses are only created if the protocols are enabled in the settings (as apps).

beckstefan: For tests you like to have an empty DB to start with. It's about not have something that could interfere. On the other hand, if you expect the DB to work with the data, tests should work if the data is present.

Technically we could bound the initial licenses to named protocols, although they are meant to be per repo---but I bet that the migration is performed before disabling the protocol in the source.

wetneb: I see your point about tests, but on the other hand I would argue that there is no point testing things in an artificial state where there are no licenses in the database. This model is introduced to replace an hard-coded list, so in all desirable testing scenarios the corresponding table will not be empty. 

Not loading the data in a migration will make our life harder in repository tests (we need to load the licenses explicitly in addition to the other objects) and when deploying an instance (we need to remember to run an extra command to populate this model).

Adding a command to load migrations manually also has the downside of not being extensible: if I want to write a new deposit interface, I will need to modify that command to add the default licenses for that repository. If we use migrations, I just need to add my licenses to a migration in my app.

beckstefan: Well, yes, that's unit test and integration test and some different philosophy here on how to do them. That's fine :-)

Let's keep the migration then.

deposit/models.py
```diff
@@ -134,6 +157,25 @@ def __str__(self):
...
+    """
+    #: FK to license
+    license = models.ForeignKey(License, on_delete=models.CASCADE)
+    #: FK to repository
+    repository = models.ForeignKey(Repository, on_delete=models.CASCADE)
+    #: Identifier used for transmission
+    transmit_id = models.CharField(max_length=255, blank=False)
+    #: True if default license for this repository
+    default = models.BooleanField(default=False)
+
```
wetneb: I wonder if we should have a field, either here or in the `License` model, to make it possible to sort licenses.
We probably want to be able to control in which order the licenses are presented to the user (for instance, creative commons licenses should probably be grouped together).

I guess we can also decide to use the primary key of this model for that. We can define the order for each repository by creating the `LicenseChooser` objects in the desired order (it should not be a problem to delete them all and recreate them, as no FK points to them, right?)

doc/sphinx/localization.rst
```diff
@@ -21,6 +21,23 @@ This lets us translate strings in various places:
...
+
+* License (field ``name``) in ``deposit.models``
+
+For localization we use `django-vinaigrette <https://pypi.org/project/django-vinaigrette/>`_. Please read their documentation for further information. In short: You have to keep in mind that:
+
+* in admin interface you do not see the localized strings,
+* you should add only english in the admin interface,
+* you will have to recreate the ``*.po`` files and add the translation manually (see below),
+* your local translations do not interact with TranslateWiki.
+
```
wetneb: Given that generating the ``.po`` files is currently handled by the developer (it is not done by Translatewiki), I guess for the production instance we should be able to generate the ``.po`` files that include the strings for the licenses of the protocols we support, right? (That does not mean we need to put that in the docs, just checking I understood the system).

deposit/migrations/0013_populate_licenses.py
```diff
@@ -0,0 +1,105 @@
+from django.db import migrations
+
```
wetneb: Don't you think it would be worth splitting this into one migration per protocol? I would add them to `deposit/zenodo/migrations` and `deposit/osf/migrations`, so that the migrations are only run if these apps are enabled in the settings. That would also help keeping repository-specific things in their own subfolders, and demonstrate to a third-party how to add support for another protocol without interferring with the generic deposit app.

deposit/models.py
beckstefan: Currently the order is hard-coded in alphabetical order in the `Protocol` class, so for now all the CCs are grouped together and it's relative stable. But having the order customized with a simple system, i.e. entering any number for each LicenseChooser object is a good idea. A possible "No license" or "Other license" as last option seems reasonable to me.

Using PK is no good. Adding license "A" means to delete and recreate alls LicenseChooser for all repos that use license A, assuming you want alphabetical order.

deposit/zenodo/protocol.py
```diff
@@ -41,7 +39,6 @@ class ZenodoProtocol(RepositoryProtocol):
     """
     A protocol to submit using the Zenodo API
     """
-    form_class = ZenodoForm
```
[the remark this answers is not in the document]
beckstefan: I agree.

doc/sphinx/localization.rst
beckstefan: For the protocol, as we discussed in chat: Yes.

deposit/migrations/0013_populate_licenses.py
beckstefan: I'll do as follows:
* Non-CC go into zenodo/osf
* LicenseChoosers go into zenodo/osf

beckstefan:
I added
* ZenodoForm
* Custom order
* Separation of migrations

(Unfortunately the commits are not exactly corresponding)

PR #769, opened 2019-05-19
Add django-extensions

wetneb:
Awesome extension which, among others, lets you generate graphs for your models:
models

Other useful things it can do:
* `./manage.py shell_plus` gives you a shell where all the models are already imported! No more `from papers.models import *`, you can directly go ahead and query stuff
* `./manage.py run_script my_script.py` lets you run a Python script which uses Django's models, without having to initialize Django yourself!

And plenty of other great things!


requirements.txt
```diff
@@ -14,6 +14,7 @@ django-capture-tag
 django-cas-ng
 django-countries
 django-crispy-forms
+django-extensions
```
Phyks: Do we want it in `requirements.txt` or rather `requirements-dev.txt`?

wetneb: Makes sense, forgot about this one.

PR #775, opened 2019-06-07
MODS for SWORDMETSProtocol

beckstefan:
Closes #774.

- [x] Prepare Testclasses
  - [x] Superclass for TestSWORDMETSProtocol
  - [x] Adapat HTML-Test
- [x] Create metadata
  - [x] DOI
  - [x] Pubytypes
    - [x] journal-article
    - [x] proceedings-article
    - [x] book chapter
    - [x] book
    - [x] journal-issue
    - [x] proceedings
    - [x] reference-entry
    - [x] poster
    - [x] report
    - [x] thesis
    - [x] dataset
    - [x] preprint
    - [x] other
- [x] Make tests derive from TestSWORDMETSProtocol
- [x] Documentation
  - [x] General documentation
  - [x] Make samples from shell
  - [x] Integrate samples in Sphinx

PR #779, opened 2019-06-26
New Start Page

beckstefan:
Content for the new start page was discussed in one of the developer meetings and PR #736.

The latest version is found here: https://annuel.framapad.org/p/Dissemin-frontpage

Basic concept is: Per screen the text is on the left and the picture on the right or vice versa. On mobile devices, the picture is always on top of the text.

- [x] Colours
  I used here a sixtone concept including the dissemin-blue. As there are more than six boxes, I repeat the colours.
  A nice offline colour chooser GPick.
- [ ] Pictures / Graphics
  They still need to be designed. TUDA can employ a professional designer, so that we just have to have some ideas. Alternatively, we use pictures.
  - [ ] Create and include pictures / graphics
  - [ ] Change alt-text
- [x] Text must be marked for translation
- [x] Include latest uploads
  - [x] Design
  - [x] Tests (Merge #775 to use its fixtures)
  - [x] Tests (Merge #796 to use its fixtures)
- [x] Align diagram in **Dissemins Data**
- [ ] Add partners
- [ ] Check font size
- [x] Make the Navigation bar smaller

This PR creates or changes translations.

~~PR fails due to #778.~~


Phyks:
Sorry, bad UI from Github, I changed the WIP status of this one by accident :/

Phyks:
A few comments on this:

- The header no longer uses the full width but is constrained to 800px. I'd rather have it use the full width.
- There is a change in the fonts weight and font size. We moved from 400 and 17.4px (or smaller) to 200 and 21px. This is too thin and less readable and we lack hierarchy (advanced search link is same size as the rest of the text for instance).
- "Panels" (green, purple, etc) have a very large top padding, similar to the "Welcome to Dissemin" screen. I think they'd rather have a small top padding as it was previously the case for the two bottom boxes.

Concerning the added features:
- I would try without margins between the blocks, this adds too much structure I think. See for instance https://kresus.org/. Same for the lateral margins, to reduce this strong block structure.
- Concerning the width of the header, I agree that it is probably better to have the header the same width as the content, but in this case it might be better if the background is truncated to a "page width" (content). Actually, it might be worth giving a try with a fully responsive layout (no fixed width).

Phyks:
Proposal after today's dev meeting:

home
papers

I'll edit the PR accordingly.

Phyks:
Should have been updated according to my latest screenshots. I'll let you double check.

I tend to think the image should be either in between title and paragraph or after the paragraph in mobile view (instead of on top of the title), but let's see when we have the actual images. No strong opinions here.

beckstefan:
Thank you!

> I tend to think the image should be either in between title and paragraph or after the paragraph in mobile view (instead of on top of the title), but let's see when we have the actual images. No strong opinions here.

The image between heading and paragraph is not possible without very much effort. 

Having the image below the text on mobile is fine for me, although I like it above more. But no strong opinion.

wetneb:
Merged in master manually.

Issue #783, opened 2019-07-01
Add DDC


beckstefan:
TUPrints needs to know DDC classification of the publication. This is likely not to be automated and hardly any metadata to find, so add a field to the form.

It should be considered to add a general metadata field, working similar to the licenses.

wetneb:
We could potentially train a classifier similar to https://haltopics.dissem.in/ for that, to prefill the field.

beckstefan:
I'll keep that in mind for later.

My idea is currently to implement something semi-general, i.e. suitable for more than one classification system.

That aside: How do these classifier handle subsets? I.e. if I have a subset of DDC to be chosen from (as it is the case for TUPrints).

wetneb:
The classifier could be trained on any classification level. For HAL we have chosen to only use the root categories, but the same system could be used with more granular categories.

beckstefan:
This means: For two different granularities I have to train twice?

wetneb:
I would say so. The machine learning model could be adapted to deal with hierarchical classifications directly, though. Surely there must be packages out there that already do that very well (I have not looked into that)

beckstefan:
Ok, I'll keep that in mind for later. 

Can you roughly estimate how many training data would be required for around 100 classes?

wetneb:
For HAL we use 50,000 documents for about a dozen classes and that is more than enough. I would guess 100,000 documents for 100 classes should give good results already.

wetneb:
By the way I am starting to migrate out of our custom-made haltopics.dissem.in into https://github.com/NatLibFi/Annif, which is well documented and maintained. It should make it easier to support other classification schemes (such as DDC) in the same instance.

PR #787, opened 2019-07-03
Enhance Metadata with Classifications

beckstefan:
This tackles #783.

For now, no machine learning is involved. For this, model is probably to flexible.

- [x] Models
- [ ] Add to form
- [ ] Add metadata
- [ ] Documentation

This PR has migrations.

This PR has translations.


beckstefan:
The intended implementation is way to big for now.

PR #796, opened 2019-07-23
Test improvements

### LINKED ISSUE #790: Add localisation to test
Currently we run our validation tests against English, as this is default.

With pytest params it is easy to run against all possible languages.

We can run a weekly, nightly CI to do that for us. In normal commits and PR we still test against English only.

This should improve the quality of translations.

All we need is merge #779 to use its fixtures, make some small changes in `.travis.yml` and tell Travis about our weekly, nightly build.

beckstefan:
This PR has several improvements on tests

- [x] Use pyflakes3 instead of pyflakes
- [x] Pytest to test pages
- [x] Set up travis cron + .travis.yml
- [x] HTML check for all available languages per Travis Cron, English still default for PR/Push
- [x] Make more tests use pytest test pages

Closes #790


beckstefan:
Thanks for the review!

Currently I have set up Travis to run the test once a week (should be on Tuesday), but TranslateWiki commits twice, which essentially makes everything double as long. Not a big deal, but the validator sometimes can't recover and won't find all errors, so it might be an idea to run Travis on daily bases.

Some kind of notification should also be nice. Otherwise someone has to look at least once a week into Travis.

PR #798, opened 2019-07-29
Patch Django-Haystack for memory consumption in queryset caching, #695

wetneb:
Hopefully this should make Dissemin a lot more efficient. Our server running django and celery currently eats up almost all of our 16GB RAM…

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,8 @@ django-cas-ng
 django-countries
 django-crispy-forms
 django-geojson
-django-haystack
+# django-haystack, currently a custom fork because of #695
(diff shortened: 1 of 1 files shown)


wetneb:
I have installed the fork in production, let's see if celery eats less memory in a few days time.

Phyks:
A new version (0.8.1) of sparse list has been deleted. We probably no longer need the fork for this one.

How is going RAM consumption (with `/search` for instance) with this PR?

wetneb:
It seems to work fine, but I recommend not to merge this until https://github.com/django-haystack/django-haystack/pull/1682 is merged. This is currently stalled by https://github.com/johnsyweb/python_sparse_list/issues/16.

Phyks:
Ok!

PR #814, opened 2019-09-04
Depositors declaration

beckstefan:
This adds a depositors declaration.

Dissemin generates after the deposit a prefilled letter, ready to be signed by the scientist.

This is needed for some repositories (by law).

- [x] Models and Admin
  - [x] Models
  - [x] Admin UI to select which letter to generate
  - [x] Store License in DepositRecord if one is chosen
  - [x] License FK in Deposit Record does need `blank=true`
- [x] Integration in UI
  - [x] View to deliver PDF (iff status is `pending`)
  - [x] Add box yelling at the user after successful deposit
  - [x] Offer Option to regenerate in "My Deposits" as long status is `pending`
  - [x] Show chosen license in "My Deposit" if one chosen
- [x] Design letter
  - [x] Register letter so that it can be selected in repository admin view
  - [x] Generate the letter for ULBDA
  - [x] Some tests
- [x] Documentation

Screenshot_2019-09-06 Short - Stefan Beck - Dissemin

Screenshot_2019-09-06 Short - Stefan Beck - Dissemin(1)

Screenshot_2019-09-06 Short - Stefan Beck - Dissemin(2)

Screenshot_2019-09-09 My deposits - Dissemin


deposit/declaration.py
```diff
@@ -0,0 +1,256 @@
...
+
+# The dict contains the available functions and an admin-friendly name
+REGISTERED_DECLARATION_FUNCTIONS = {
+    'declaration_ulb_darmstadt' : 'ULB Darmstadt',
+}
+
+
+def get_declaration_pdf(deposit_record, user):
+    """
+    This function creates for a given deposit the letter of declaration. If the does not succeed, it raises an exception.
```
Phyks: "the" should probably be "this".

deposit/declaration.py
```diff
@@ -0,0 +1,256 @@
...
+    return func(deposit_record, user)
+
+
+def get_declaration_function(func_name):
+    """
+    This function gets the function to generate a letter of declaration. If no function can be found, we raise AttributeError.
+    :param func_name: Name of the function that is requested
+    :returns: function or raises AttributeError
+    """
+    return getattr(current_module, func_name)
```
Phyks: I personally don't like this much. Why not listing the declaration functions directly within the `REGISTERED_DECLARATION_FUNCTIONS` dict?

deposit/declaration.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+def declaration_ulb_darmstadt(deposit_record, user):
+    """
+    Takes a deposit and creates authors declaration for ULB Darmstadt and returns that.
+    This function follows the corporate design of TUDA.
+    As this is my first approach with reportlab, the way placing things is not really good
```
Phyks: "the way things are placed"

deposit/declaration.py
```diff
@@ -0,0 +1,256 @@
...
+    <br/>
+    (Autoren- und Herausgebervertrag)
+    """
+    available_width = page_width - 30*mm
+    available_height = 30
+    subject_p = Paragraph(subject_text, style_subject)
+    used_width, used_height = subject_p.wrap(available_width, available_height)
+    subject_p.drawOn(c, 15*mm, page_height - 74.4*mm - used_height)
+
+    # Main text of the declaration. It's top is going the an arbitrary set point and remaining height from dynamic generated content is referenced to this height
```
Phyks: Second sentence is not super clear :)

deposit/migrations/0018_letter_of_declaration.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.2.4 on 2019-09-06 13:52
+
+from django.db import migrations, models
+import django.db.models.deletion
```
Phyks: `from` and `import` should be swapped.

deposit/views.py
```diff
@@ -246,3 +260,26 @@ def submitDeposit(request, pk):
...
+        """
+        dr = get_object_or_404(DepositRecord.objects.select_related('paper', 'repository', 'user', 'license'), pk=pk)
+
+        if dr.user != request.user:
+            return HttpResponseForbidden(_("Access to this ressource not allowed."))
+        # If the repository requires a letter of declaration, we try to create the pdf, otherwise we return 404.
+        if dr.repository.letter_declaration != '' and dr.status == "pending":
+            pdf = get_declaration_pdf(dr, request.user)
+            pdf.seek(0)
+            filename = "Erklärung {}.pdf".format(dr.paper.title)
```
Phyks: filename should probably be localized.

papers/views.py
```diff
@@ -500,20 +500,25 @@ def get_object(self, queryset=None):
...
+                if dr.paper_id == self.object.id and dr.status in ['pending', 'published']:
+                    if self.request.user.is_authenticated and self.request.user == dr.user and dr.repository.letter_declaration:
+                        dr.letter = True
+                    context['deposit'] = dr
             except (TypeError, ValueError, DepositRecord.DoesNotExist):
                 pass
-        context['can_be_deposited'] = (not self.request.user.is_authenticated
-                    or self.object.can_be_deposited(self.request.user))
+
+        context['can_be_deposited'] = (not self.request.user.is_authenticated or self.object.can_be_deposited(self.request.user))
```
Phyks: Would maybe be better as 

```
context['can_be_deposited'] = (
    not self.request.user.is_authenticated or
    self.object.can_be_deposited(self.request.user)
)
```

papers/views.py
```diff
@@ -500,20 +500,25 @@ def get_object(self, queryset=None):
...
+
+        context['deposit'] = None
         if 'deposit' in self.request.GET:
             try:
                 pk = int(self.request.GET['deposit'])
-                dep = DepositRecord.objects.get(pk=pk)
-                if dep.paper_id == self.object.id:
-                    context['deposit'] = dep
+                dr = DepositRecord.objects.select_related('repository', 'oairecord').get(pk=pk)
+                print(dr.status)
```
Phyks: Forgotten print instruction

deposit/declaration.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+def declaration_ulb_darmstadt(deposit_record, user):
+    """
+    Takes a deposit and creates authors declaration for ULB Darmstadt and returns that.
+    This function follows the corporate design of TUDA.
+    As this is my first approach with reportlab, the way placing things is not really good
+    """
+
+    pdf_buffer = io.BytesIO()
```
Phyks: This approach of generating the PDF file from scratch with `reportlab` in Python is definitely fine for me. Not sure though, but would it be possible to have instead a source PDF file with forms and just pre-fill the forms in Dissemin?

My idea behind this is that it would be easier to add new repositories since the contributor would just have to contribute a PDF file with forms, which can be easily generated from any PDF creation software.

Definitely not a blocking point for merging though.

deposit/declaration.py
beckstefan: Fixed, thanks.

deposit/declaration.py
beckstefan: Fixed, thanks

deposit/declaration.py
beckstefan: Should be clearer now

deposit/migrations/0018_letter_of_declaration.py
beckstefan: Yes, thanks!

deposit/declaration.py
beckstefan: Yes, I have thought about that, too. I prefer this way to, but it seemed as the more difficult approach to me.

I'll do some investigation on that and open an issue.

papers/views.py
beckstefan: Removed, thanks

papers/views.py
beckstefan: Changed, thanks!

deposit/views.py
beckstefan: At it least it won't hurt :)

deposit/declaration.py
beckstefan: I was thinking the wrong way and was (once again) not aware of how awesome python can be.

Thank you!

PR #827, opened 2019-09-30
Fix minor issues with the start page

Phyks:
* Header was too large on mobile device
* Partners images were not readable on green background

### DIFF
--- a/papers/static/style/mobile.css
+++ b/papers/static/style/mobile.css
@@ -1,6 +1,9 @@
 /* small screen fixes */
 
-#content { min-width: 300px; }
+
+#header, #content, #subfooter, #footer {
+    min-width: 300px;
+}
 
 #header .navbar .navbar-nav {
     flex-direction: column;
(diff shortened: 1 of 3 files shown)


beckstefan:
> Partners images were not readable on green background

i am slightly irritated as I could read them quite well.

Nevertheless: Thanks!

Phyks:
> i am slightly irritated as I could read them quite well.

I meant this, depending on your screen, it was a bit difficult to read (especially the base one, on my mobile phone).

Screenshot_2019-09-30 Home - Dissemin

But my main concern is actually that there should probably be some styling guide somewhere from these institutions saying that their logos should not be displayed on anything else than a white background. https://www.inria.fr/en/content/view/full/11228 for instance (for Inria, not a partner)

beckstefan:
> But my main concern is actually that there should probably be some styling guide somewhere from these institutions saying that their logos should not be displayed on anything else than a white background.

Yes. In case I would wonder, why they then have logos with transparency. Below is some quick research.

* BASE: No information on https://www.base-search.net/about/de/about_infomat.php
  We could directly the white background

* OpenIng: No restriction

* ENS-PSL: No information on https://www.ens.fr/logotype-de-l-ens

* Couperin: No transparency, no issue

I see now the issue, so yes, white boxes are nice. Thanks!

PR #830, opened 2019-10-07
Sword mets improvements

beckstefan:
This improves the SWORD METS protocol

* Provide a small header in mets.xml which states that the metadata is generated by dissemin
* Centralize namespaces in mets.xml in root element
* Provide MODS examples as bundle


wetneb:
Hi Stefan,

Thanks for this! I am not in a position to review pull requests as I no longer have time to work on dissemin and I am not very up to date on the latest developments. Sorry about that!

PR #833, opened 2019-10-09
Enables non-interactive frontend for APT

rgrunbla:
Hey,

trying to use vagrant to install dissemin, I ran into one issue with grub breaking the workflow (basically, some ncurses menu rendered as text during the installation). This was due to APT not being configured to use a non-interactive installation method, which this pull requests enable.

Rémy


beckstefan:
Although I do not use vagrant (and not being experienced with it), I propose to merge. 

@Phyks @wetneb What do you think?

Phyks:
Good for me

Issue #874, opened 2020-02-19
GeoJSON


beckstefan:
Currently we have a dependency on *django-geojson*, that is not really used in our code, as far as I can see.

*django-geojson* broke our build, because it has a dependency on *jsonfield*. The latter one has recently had a major update, so that *django-geojson* fails with ``django.core.exceptions.ImproperlyConfigured: `jsonfield` dependency missing. See README.``

So maybe, we can remove the functionality?

(Currently, I have set out dependency on *jsonfield* to ``<3``.)

wetneb:
Yes sure! It was added for the map of institutions but that has never been finished.

PR #876, opened 2020-02-27
Remove django-geojson pt 1

beckstefan:
This is the first of two PR to solve #874 

First, we need to remove the field, which requires django-geojson to be installed. This needs to be deployed.

Second, we can replace migrations 39 to 59 by a squashed version, that does not depend on django-geojson.


beckstefan:
Definitely yes and I am not going to delete migrations 39-59 very soon. This is the main reason, why it's two parted.

wetneb:
Oh right, the ones you are deleting have been squashed a long time ago!

PR #897, opened 2020-10-08
Hide Oairecords if they have a deposit that is not available

beckstefan:
Currently we even show links/sources, where we have status 'rejected'. A
lot of pending records point to 404 pages, which is after all very
misleading.
There's still a box for pending or embargoed records on the paper detail
page.

### DIFF
--- a/papers/models.py
+++ b/papers/models.py
@@ -68,6 +68,7 @@ from django.core.exceptions import ObjectDoesNotExist
 from django.urls import reverse
 from django.db import DataError
 from django.db import models
+from django.db.models import prefetch_related_objects
 from django.template.defaultfilters import slugify
 from django.utils import timezone
 from django.utils.functional import cached_property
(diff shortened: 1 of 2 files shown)


papers/models.py
```diff
@@ -697,6 +697,14 @@ def cache_oairecords(self):
         """
         self.cached_oairecords = list(self.oairecord_set.all())
 
+    @property
+    def sorted_oairecords_with_published_deposit(self):
+        """
+        Fetches all oairecords; if there's a DepositRecord, require that it is published
+        """
+        l = [r for r in self.sorted_oai_records if not r.depositrecord_set.exists() or r.depositrecord_set.filter(status='published')]
```
wetneb, line 705: That is likely to generate a lot of SQL requests, no? Perhaps it'd be more efficient to use the visibility field of the OaiRecord model: make sure we mark them as invisible when we create OaiRecords that are not published yet, and update the visibility once the deposit succeeds. That will also require a migration to set the visibility on existing pending deposits.

wetneb (commented):
Thanks a lot for tackling this!

papers/models.py
beckstefan: Yes, it does. Alternatively I can design a single query with `prefetch_related` and not using `self.sorted_oai_records`.

Visibility is a field of Paper model, not of OaiRecord model. It was my first intention, too.

wetneb: Ah makes sense. I don't know if it is worth adding a field on OaiRecord for that, then. Perhaps `prefetch_related` is good enough in that case then.

beckstefan: Prefetch related does two queries anyway, so we can still uses cached sorted oairecords and fetch the DepositRecords for them after.

wetneb (approved):

